PR #3, opened 2019-08-27
fix url_keys

snackmgmg:
README.md says, 

```
SELECT bqutil.fn.url_keys(
  'https://www.google.com/search?q=bigquery+udf&client=chrome')

["q", "client"]
```

But I get,

```
["https://www.google.com/search?q","client"]
```

I think this result is unexpected. So I fixed it.


snackmgmg:
thanks!

PR #7, opened 2019-09-02
Data access work

mescanne:
Sample SQL script for analysing data access logs produced by GCS and BigQuery datasets.

I've adjusted README.md to add it in and provided instructions for setting up the sink, data access controls, as well as running the SQL.

### DIFF
--- a/scripts/data_access/create_data_access_view.sql
+++ b/scripts/data_access/create_data_access_view.sql
@@ -1,0 +1,132 @@
+#standardSQL
+
+--
+-- This is setup for storage.googleapis.com and bigquery.googleapis.com
+--
+-- This is for examining access to bigquery datasets and buckets.
+-- 
+-- Table and object-level is available in the logs (and more!), but
+-- this creates confusion for a dashboard.
+-- 
+CREATE OR REPLACE VIEW data_access.audit_summary AS
+WITH
+  -- Pull out Data Access logs
+  DataAccess AS (
+    SELECT
+      -- Hour truncated
+      TIMESTAMP_TRUNC(d.timestamp, HOUR) AS hour,
+      -- Project ID that the access method was called on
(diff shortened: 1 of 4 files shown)


scripts/data_access/create_data_access_view.sql
```diff
@@ -0,0 +1,132 @@
...
+      ELSE
+        CONCAT('Unknown bigquery:', action)
+      END
+    ELSE
+      CONCAT('Unknown service:', service)
+  END AS op,
+  granted,
+  -- Project is of the resource or, if not there,
+  -- then for the method accessing it (eg for buckets)
+  CASE
```
freedomofnet: Do you need a ELSE to set a default value here? Otherwise, it will create NULLs.

scripts/data_access/create_data_access_view.sql
```diff
@@ -0,0 +1,132 @@
...
+      d.protopayload_auditlog.serviceName IN ('storage.googleapis.com',
+                                              'bigquery.googleapis.com')
+  )
+SELECT
+  hour,
+  service,
+  actor,
+  -- Translate the action into an operation (READ/WRITE/ADMIN)
+  CASE
+    WHEN service = 'storage' THEN
```
freedomofnet: Do you think by checking the action value, you can determine the action type? So, maybe simplify the code a little with one level of CASE statement?

Same for service='bigquery' code below.

scripts/data_access/data_access_policy.json
```diff
@@ -0,0 +1,8 @@
+{
```
freedomofnet: Please add a description, what this file is for?

freedomofnet:
Hi mescanne, 

Please take a look of the comments when you have time.

Thanks!

PR #11, opened 2019-09-21
implement teradata left and right functions

hashkanna:
This implements the [LEFT](https://docs.teradata.com/reader/1DcoER_KpnGTfgPinRAFUw/Ef3DAn9emz~W57YSCkoEOQ) and [RIGHT](https://docs.teradata.com/reader/1DcoER_KpnGTfgPinRAFUw/4bAyNUOB6_VOuvSVrDRqDQ) functions of Teradata

### DIFF
--- a/udfs/migration/teradata/left.sql
+++ b/udfs/migration/teradata/left.sql
@@ -15,5 +15,8 @@
  */
 
 CREATE OR REPLACE FUNCTION td.left(string_expr STRING, length_expr INT64) AS (
-  SUBSTR(string_expr, 1, length_expr)
+  CASE
+    WHEN length_expr < 0 THEN ERROR('The length argument to LEFT must be a positive integer!')
+    ELSE SUBSTR(string_expr, 1, length_expr)
+  END
 );
--- a/udfs/migration/teradata/left.sql
+++ b/udfs/migration/teradata/left.sql
@@ -1,0 +1,19 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE OR REPLACE FUNCTION td.left(string_expr STRING, length_expr INT64) AS (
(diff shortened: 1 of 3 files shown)


udfs/migration/teradata/right.sql
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE OR REPLACE FUNCTION td.right(string_expr STRING, length_expr INT64) AS (
+  SUBSTR(string_expr, LENGTH(string_expr) - length_expr + 1, length_expr)
```
ryanmcdowell: The value of `LENGTH(string_expr) - length_expr + 1` will become negative if the `length_expr` is  > the length of the `string_expr`. This causes the `SUBSTR` function to begin truncating from the end of the string again which is unexpected behavior. Perhaps the following would be better.

```sql
CASE
  WHEN length_expr < 0 THEN ERROR('The length argument to RIGHT must be a positive integer!')
  ELSE SUBSTR(string_expr, -1 * length_expr, length_expr)
END
```

udfs/migration/teradata/right.sql
```diff
@@ -0,0 +1,19 @@
...
+);
```
ryanmcdowell: Add newline character to the end of file.

udfs/migration/teradata/left.sql
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE OR REPLACE FUNCTION td.left(string_expr STRING, length_expr INT64) AS (
+  SUBSTR(string_expr, 1, length_expr)
+);
```
ryanmcdowell: Add newline character to the end of the file.

udfs/migration/teradata/left.sql
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE OR REPLACE FUNCTION td.left(string_expr STRING, length_expr INT64) AS (
+  SUBSTR(string_expr, 1, length_expr)
```
ryanmcdowell: I would handle the negative-case because `SUBSTR` will throw an error with a negative `length_expr` and it may be hard for the user to debug.

PR #16, opened 2019-10-13
Add Redshift INITCAP function

hashkanna:
Implement Redshift [INITCAP](https://docs.aws.amazon.com/redshift/latest/dg/r_INITCAP.html) function by making use of re2 [Unicode character class names \p{P} and \p{S}](https://github.com/google/re2/wiki/Syntax), and [order of regex alternatives](https://superuser.com/questions/1194399/does-order-of-alternatives-in-regex-matters)

### DIFF
--- a/udfs/migration/redshift/initcap.sql
+++ b/udfs/migration/redshift/initcap.sql
@@ -15,16 +15,18 @@
  */
 
 CREATE OR REPLACE FUNCTION rs.initcap(string_expr STRING) AS (
-  (SELECT
-    CASE WHEN string_expr = '' THEN ''
-    ELSE STRING_AGG(
-      CONCAT(
-        UPPER(SUBSTR(word,1,1))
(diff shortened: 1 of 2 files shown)


udfs/migration/redshift/initcap.sql
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE OR REPLACE FUNCTION rs.initcap(string_expr STRING) AS (
+  (SELECT
```
danieldeleo: Remove the unnecessary parenthesis around the entire SELECT statement

PR #25, opened 2019-11-18
Zero-norm a variable

lakshmanok:


udfs/community/README.md
```diff
@@ -23,6 +23,7 @@ SELECT bqutil.fn.int(1.684)
 * [translate](#translateexpression-string-characters_to_replace-string-characters_to_substitute-string)
 * [url_keys](#url_keysquery-string)
 * [url_param](#url_paramquery-string-p-string)
+* [zero_norm](#zero_norm)
```
freedomofnet: Since your function name is zeronorm, please keep it consistent.

udfs/community/README.md
```diff
@@ -153,3 +154,28 @@ SELECT bqutil.fn.y4md_to_date('20201220')
 
 "2020-12-20"
 ```
+
+### [zeronorm(x ANY TYPE, meanx FLOAT64, stddevx FLOAT64)](median.sql)
```
ryanmcdowell: Change median.sql to zeronorm.sql

udfs/community/README.md
ryanmcdowell: Also, in order for GitHub to properly link it, the ref link should be:
```
#zeronormx-any-type-meanx-float64-stddevx-float64
```

fhoffa:
Wondering if it would be more useful if you pass it an ARRAY_AGG() of the variable, then compute AVG() and STDDEV() inside. Depends mostly on your context!

lakshmanok:
I don't think it's do-able because you can not do x - avg(x),   Two
reasons:  (1) AVG() etc. are not defined for arrays (2) BigQuery doesn't
support broadcasting a single element to the length of an array.

PR #34, opened 2019-12-16
Create stored_procedure get_next_ids.sql

freedomofnet:
It generates next ids and inserts them into a sample table. This implementation prevents against race condition.


freedomofnet:
The README is already in the stored_procedures folder, and the usage example is there also. Thanks.

PR #36, opened 2020-02-08
Adding UDF Testing Framework

danieldeleo:
Continuation of previous PR: https://github.com/GoogleCloudPlatform/bigquery-utils/pull/21


udfs/CONTRIBUTING.md
```diff
@@ -0,0 +1,92 @@
...
+functions work as expected. So testing these functions locally before submitting
+the pull request can ensure a successful review process.
+
+## Testing UDFs Locally
+
+Please follow these instructions to confirm that the test cases being provided
+in your pull request work as expected.
+
+1.  Change into the bigquery_utils top-level directory.
+1.  Create a Python virtual environment and activate it:
```
jaketf: This venv is bare. Are you missing a step to pip install dependencies?

udfs/CONTRIBUTING.md
```diff
@@ -0,0 +1,92 @@
...
+
+1.  Add your UDFs (**one** UDF per file) and a
+    [Contributor License Agreement](#contributor-license-agreement) to the
+    appropriate directory.
+    *   If your function replicates logic from some other data warehouse UDF,
+        place it in the relevant sub-directory in the
+        [migration](/udfs/migration) directory. Otherwise, place it in the
+        [community](/udfs/community) directory.
+1.  Add test cases for your UDFs.
+    *   Edit the `test_cases.yaml` file to include test inputs and expected
```
jaketf: nit: can we relative link to this file?

udfs/CONTRIBUTING.md
```diff
@@ -0,0 +1,92 @@
...
+        outputs for the function.
+    *   Make sure test cases provide full coverage of the function's expected
+        behavior. For example, if integers are the expected input, please
+        provide test cases with the following inputs: negative numbers, zero,
+        positive numbers, and null values.
+1.  Describe what your UDF does.
+    *   Edit the `README.md` in the associated sub-directory to include a
+        description of the function and make sure your function is listed in
+        alphabetical order amongst the other functions in the `README.md`.
+    *   Make sure the same description is placed as a comment in your UDF file.
```
jaketf: nit / nice to have in future PR:
1) alphabetizing is annoying and error prone (speaking for dyslexic experience)
2) Keeping comments / README in sync is also annoying / error prone.

Solution:
DRY up documentation by auto-generate these READMEs instead of having this burden on contributor.
Much like javadoc or sphinx or terraform-docs These udf  READMEs should generated from the the comment in the UDF SQL files.

udfs/community/test_cases.yaml
```diff
@@ -0,0 +1,62 @@
+int:
```
jaketf: Should this file be split up per function?
Seems like this could get very long as number of UDFs grows.

udfs/tests/udf_test_utils.py
```diff
@@ -0,0 +1,122 @@
...
+    udf_dir = '/'.join(udf_path.split('/')[:-1])
+    yaml_test_data_path = f'{udf_dir}/test_cases.yaml'
+    if path.isfile(yaml_test_data_path):
+        with open(yaml_test_data_path, 'r') as yaml_file:
+            return load(yaml_file, Loader=SafeLoader)
+    else:
+        return None
+
+
+def extract_udf_name(udf_path):
```
jaketf: nit: test-ception
This logic in your testing framework utilities could use unittests.

udfs/CONTRIBUTING.md
```diff
@@ -0,0 +1,92 @@
...
+
+Please follow these instructions to confirm that the test cases being provided
+in your pull request work as expected.
+
+1.  Change into the bigquery_utils top-level directory.
+1.  Create a Python virtual environment and activate it:
+    *   `python3 -m venv venv`
+    *   `source venv/bin/activate``
+1.  Run all tests by invoking the `run.sh` script
+    *   `bash udfs/tests/run.sh`
```
jaketf: This should mention the aruments `--[create/delete]_datasets` arguments.
We should call out that testing "locally" requires creating BQ datasets in your own project to facilitate the tests.

udfs/tests/run.sh
```diff
@@ -0,0 +1,29 @@
...
+# limitations under the License.
+
+#########################################################################
+# Call run.sh locally with an optional parameter specifying the test name
+# if you want to run only that test and no others
+#
+# e.g. bash udfs/tests/run.sh test_community_udf
+#
+#########################################################################
+python3 -m pip install -r udfs/tests/requirements.txt
```
jaketf: Installation of dependencies should happen outside of this script so the installation of dependencies doesn't happen on every test run.
locally I shouldn't have to run this pip step every time I want to run the tests just to functionally no-op.

I think you did this so you can just use the python slim image directly.

It would be ideal to use the Docker cloud builder to build an image (Dockerfile base image python slim + this pip install line) or pull from cache in gcr that has these deps installed and the cloud build should just use this as your base image. We could make this publicly available gcr image for the purpose of contributing to this repo. This [example](https://github.com/GoogleCloudPlatform/professional-services/blob/master/examples/cloud-composer-cicd/cloudbuild.yaml) from professional-services might be helpful

If you don't want to manage the container in GCR or want to be lazy for the purposes of this PR, then at least move this line into the cloud build.
For Example
```
- name: python:3.8.0-slim
  entrypoint: 'bash'
 args: ["python3 -m pip install -r udfs/tests/requirements.txt && udfs/tests/run.sh"]
```

Regardless we should add to contributing a step to install the dependencies in the virtualenv so for local testing.

udfs/tests/udf_test_utils.py
danieldeleo: Addressed implicitly by adding a test to ensure that extract_udf_name is equivalent to the file name

udfs/migration/teradata/otranslate.sql
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-CREATE OR REPLACE FUNCTION td.translate(
+CREATE OR REPLACE FUNCTION td.otranslate(
```
jaketf, line 17: is this intentional?

danieldeleo: yes intentional. I'm fixing the name since this function is supposed to be otranslate as the author wrote in README https://github.com/GoogleCloudPlatform/bigquery-utils/tree/master/udfs/migration/teradata#otranslatesource_string-string-from_string-string-to_string-string

udfs/tests/udf_test_utils.py
```diff
@@ -0,0 +1,122 @@
...
+        f'`{project_id}.\\g<bq_dataset>_test\\g<udf_name>`(',
+        udf_sql)
+    if udf_length_before_replacement == len(udf_sql):
+        return None
+    else:
+        return udf_sql
+
+
+def get_target_bq_dataset(udf_path):
+    parent_dir_name = str(dirname(udf_path).split('/')[-1])
```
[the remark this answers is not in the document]
danieldeleo: REALLY excited to join the cool kids, awesome tip, thanks!

PR #57, opened 2020-05-07
Add json_typeof

apstndb:
Add [`json_typeof` function](https://www.postgresql.org/docs/12/functions-json.html) of PostgreSQL.
It seems [not available in RedShift](https://docs.aws.amazon.com/redshift/latest/dg/json-functions.html) so I don't place it into `udfs/migration/redshift`.


apstndb:
The test seems broken because `json` has `js`.
https://github.com/GoogleCloudPlatform/bigquery-utils/blob/1b6fc90eb0a434e36fdc7957fe030b7e622ddb6c/udfs/tests/udf_test_utils.py#L70

ryanmcdowell:
> The test seems broken because string `json` contains `js`.
> https://github.com/GoogleCloudPlatform/bigquery-utils/blob/1b6fc90eb0a434e36fdc7957fe030b7e622ddb6c/udfs/tests/udf_test_utils.py#L70

Fixed in #59

PR #58, opened 2020-05-07
Calculate percentage difference between two values

christippett:
This is a calculation I use quite often and make use of as a UDF. I thought it'd be useful for others too, hence the PR.


christippett:
@googlebot I signed it!

udfs/community/percentage_difference.sql
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE OR REPLACE FUNCTION fn.percentage_difference(val1 NUMERIC, val2 NUMERIC) AS (
+  CASE WHEN val1 > 0 AND val2 > 0 THEN 
+    ABS(val1 - val2) / ((val1 + val2) / 2)
+  END
+);
```
ryanmcdowell, line 21: Add trailing newline

christippett:
Thanks for the feedback @ryanmcdowell.

I've added test cases that cover different data type input scenarios, as well as a few others. I've also updated the README.

Note: I've rounded the output with an opinionated precision of `4`, which I think is reasonable for a percentage value - especially as the output is a FLOAT64 value. What do you think?

PR #63, opened 2020-05-18
Implemented Snowflake factorial function

epishova:
Added snowflake dataset and implemented sf.factorial function


README.md
```diff
@@ -25,6 +25,7 @@ This repository is broken up into:
         *   [redshift](/udfs/migration/redshift)
         *   [teradata](/udfs/migration/teradata)
         *   [vertica](/udfs/migration/vertica)
+        *   [snowflake](/udfs/migration/snowflake)
```
ryanmcdowell, line 28: Alpha order

release/build.sh
```diff
@@ -58,6 +58,8 @@ function get_dataset() {
       dataset="td";;
     *"vertica"*)
       dataset="ve";;
+    *"snowflake"*)
```
ryanmcdowell, line 61: Alpha order

udfs/migration/snowflake/factorial.sql
```diff
@@ -0,0 +1,54 @@
...
+    121645100408832000,
+    2432902008176640000,
+    51090942171709440000.,
+    1124000727777607680000.,
+    25852016738884976640000.,
+    620448401733239439360000.,
+    15511210043330985984000000.,
+    403291461126605635584000000.,
+    10888869450418352160768000000.][OFFSET(integer_expr)] 
+));
```
ryanmcdowell, line 54: Add trailing newline

udfs/migration/snowflake/factorial.sql
```diff
@@ -0,0 +1,54 @@
...
+    6402373705728000,
+    121645100408832000,
+    2432902008176640000,
+    51090942171709440000.,
+    1124000727777607680000.,
+    25852016738884976640000.,
+    620448401733239439360000.,
+    15511210043330985984000000.,
+    403291461126605635584000000.,
+    10888869450418352160768000000.][OFFSET(integer_expr)]
```
ryanmcdowell, line 53: Why do these values have periods at the end?

udfs/migration/snowflake/factorial.sql
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- factorial:
+-- Computes the factorial of its input. The input argument must be an integer expression in the range of 0 to 27. 
+-- Due to data type differences, the maximum input value in BigQuery is smaller than in Snowflake. 
+-- Input:
+-- integer_expr: INT64
+-- Output: NUMERIC
```
ryanmcdowell, line 22: Looks like the Snowflake factorial function returns an integer, is there a reason we return NUMERIC?

udfs/migration/snowflake/factorial.sql
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2019 Google LLC
```
ryanmcdowell, line 2: 2020

udfs/migration/snowflake/test_cases.yaml
```diff
@@ -0,0 +1,14 @@
...
+    expected_output: 1.0
+  - test:
+    input: 1
+    expected_output: 1.0
+  - test:
+    input: 5
+    expected_output: 120
+  - test:
+    input: 10
+    expected_output: 3628800
```
ryanmcdowell, line 14: Should these expected outputs also have a `.0` since the return type is NUMERIC?

udfs/tests/udf_test_utils.py
```diff
@@ -28,6 +28,7 @@
     'teradata': 'td_test',
     'vertica': 've_test',
     'community': 'fn_test',
+    'snowflake': 'sf_test',
```
ryanmcdowell, line 31: Alpha order

udfs/migration/snowflake/test_cases.yaml
epishova: BQ does type conversion and the test passes. I can add `.0` to emphasize that the return type is NUMERIC

udfs/migration/snowflake/factorial.sql
epishova: If we use INT64, the max returned value would be `20!`. The larger INT64 values produce an overflow error. With NUMERIC the max value is `27!` which is a bit better (however still smaller than Snowflake's `33!`).

ryanmcdowell: Got it. 

nit: Can we add a trailing 0 to each value. Example: `1.0`

epishova: Added trailing zeros.

PR #68, opened 2020-06-09
Setup SQL extraction project

creativitRy:
Initial setup for the SQL extraction project, using Gradle to build Kotlin and Antlr files.


tools/sql_extraction/src/main/kotlin/App.kt
```diff
@@ -0,0 +1,9 @@
+package com.google.cloud.sqlextraction
```
feiling, line 1: Have you checked with Zihao to see if this the namespace to use?

creativitRy: Ah, I missed sqlecosystem. I can fix that.

PR #69, opened 2020-06-11
SQL Extractor: java examples, cli, file list expander and unit tests

creativitRy:
- added `examples` directory with java files to use as example inputs
- added a command line arguments library, with current options following:
```
Usage: sql_extraction [OPTIONS] [FILEPATHS]...

Options:
  -r, --recursive  scan files in subdirectories recursively
  -h, --help       Show this message and exit

Arguments:
  FILEPATHS  file and directory paths to code
```
- added basic file list expander (converts directories to list of files) and unit tests

Support for features such as globs/file extension support will be added in a future commit

### DIFF
--- a/tools/sql_extraction/build.gradle.kts
+++ b/tools/sql_extraction/build.gradle.kts
@@ -8,8 +8,12 @@ plugins {
 group = "com.google.cloud.sqlecosystem"
 version = "1.0"
 
+// clean task = clean
+// build task = build
+// build and run task = run
+// test task = test
 application {
-    mainClassName = "com.google.cloud.sqlecosystem.sqlextraction.AppKt"
+    mainClassName = "com.google.cloud.sqlecosystem.sqlextraction.CliKt"
 }
 
 repositories {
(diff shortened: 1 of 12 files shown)


tools/sql_extraction/src/main/kotlin/Cli.kt
```diff
@@ -0,0 +1,30 @@
...
+        canBeFile = true
+    ).multiple()
+
+    override fun run() {
+        logger.debug("Starting SQL Extraction from command line")
+
+        val files = FileListExpander().expandAndFilter(filePaths, recursive)
+        logger.debug { "Files to analyze: $files" }
+    }
+}
```
feiling, line 30: Please print out the usage when -h is specified. In the usage, please also include several sample usages. 

It seems right now the program doesn't support file patterns like "*.cs" and it's hard-coded to look for java files?

tools/sql_extraction/build.gradle.kts
```diff
@@ -31,6 +38,7 @@ task<de.undercouch.gradle.tasks.download.Download>("downloadGrammars") {
     }
     src(listOf("https://raw.githubusercontent.com/antlr/grammars-v4/master/java/java9/Java9.g4"))
     dest("gen/main/antlr/")
+    overwrite(false)
     println(System.getProperty("user.dir"))
 }
```
feiling, line 44: Please add some comments to list the commands on
- how to build
- how to run unit tests

tools/sql_extraction/src/main/kotlin/Cli.kt
creativitRy: Good idea for printing sample usages!

Currently, only java files are included since I wanted to focus on it first. But maybe it's a better idea to accept all file types here and filter them later based on which language frontends are implemented. This should also bring back the reason to support the --filetype flag that you mentioned.

feiling: you don't have to add support for non-java files for now.

PR #70, opened 2020-06-17
SQL Extraction: setup data pipeline (frontends, json output)

creativitRy:
To analyze each file, both a frontend and a data-flow engine are required. The frontend reads through the language-specific AST in program order and calls necessary methods in the engine (e.g. foundLiteral, foundUsage, etc). The AST parser and visitor are provided by Antlr. Since this commit was getting massive, the engine implementation and other details will be on a future commit.

Resulting list of queries are converted to JSON and printed to STDOUT.

### DIFF
--- a/tools/sql_extraction/src/main/kotlin/Cli.kt
+++ b/tools/sql_extraction/src/main/kotlin/Cli.kt
@@ -7,17 +7,31 @@ import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.multiple
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.types.path
+import com.google.gson.GsonBuilder
 import mu.KotlinLogging
 import java.nio.file.Path
 
-private val logger = KotlinLogging.logger { }
+private val LOGGER = KotlinLogging.logger { }
 
-fun main(args: Array<String>) = Cli().main(args)
+fun main(args: Array<String>) = Cli(
+    FileListExpander(),
+    SqlExtractor(
+        DataFlowSolver(
+            listOf(
+                JavaFrontEnd()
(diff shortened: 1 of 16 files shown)


tools/sql_extraction/src/main/kotlin/Cli.kt
```diff
@@ -29,31 +43,51 @@ class Cli : CliktCommand(
...
     ```
     """
-) {
-    val recursive: Boolean by option(
-        "-R", "-r", "--recursive",
-        help = "scan files in subdirectories recursively"
-    ).flag()
-    val filePaths: List<Path> by argument(help = "file and directory paths to analyze").path(
+    ) {
+    private val filePaths: List<Path> by argument(
```
feiling, line 47: Curious to know what's the different between "by argument" and "by option"?

tools/sql_extraction/src/main/kotlin/FileListExpander.kt
```diff
@@ -4,7 +4,20 @@ import java.nio.file.Files
...
 class FileListExpander {
+    /**
+     * Converts a collection of file and directory paths to all applicable file paths
+     *
+     * @param[dirs] File and directory paths to consider
+     * @param[recursive] True of visit subdirectories recursively,
+     *     false to only visit the top-level directories
+     * @param[includes] List of GLOB patterns to limit added output file paths
+     * @param[excludes] List of GLOB patterns to filter out output file paths
+     * @return Expanded and filtered collection of file paths to analyze
```
feiling, line 19: This is OK today. But pls consider what happens when there are many files to be returned. It's better in that case to use lazy evaluation and return the result in a stream.

creativitRy: Wow good idea!

PR #71, opened 2020-06-22
SQL Crawler: Added crawler architecture and files

noah-kuo:


tools/unsupervised_dataset/Crawler/Crawler.py
```diff
@@ -0,0 +1,101 @@
...
+        """ Initializes the crawler and instance variables.
+        
+        Args:
+            links: The root URLs to begin crawling from. Can be one or more.
+            maxDepth: The maximum depth for the crawler to explore.
+            maxSize: THe maximum number of links the crawler should explore.
+    
+        """
+        self.linkQueue = queue.Queue()
+        self.seen = set([])
```
yzhvictor, line 23: This can just be set()

tools/unsupervised_dataset/Crawler/Crawler.py
```diff
@@ -0,0 +1,101 @@
...
+            self.seen.add(link)
+        
+    def crawl(self):
+        """ Begins the crawling process using variables set earlier. Extracts
+            queries by locating website-specific HTML tags or searching for
+            common expression patterns. Writes queries to output after finishing
+            each site.
+        """
+        
+        while (self.linkQueue.empty() == False):
```
yzhvictor, line 40: not self.linkQueue.empty()

tools/unsupervised_dataset/Crawler/Crawler.py
```diff
@@ -0,0 +1,101 @@
...
+            # Retrieve the next link in the queue
+            nextNode = self.linkQueue.get()
+            nodeURL = nextNode.getURL()
+            nodeDepth = nextNode.getDepth()
+            
+            # Check if crawler has exceeded maximum depth or maximum count
+            if nodeDepth >= self.maxDepth or self.count >= self.maxSize:
+                print("Crawled {0} websites.".format(self.count))
```
yzhvictor, line 48: Instead of printing to console, try to print to some logger, such as https://docs.python.org/2/library/logging.html

tools/unsupervised_dataset/Crawler/CrawlerLog.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self):
+        """ Initializes crawler log to keep track of crawler progress and instantiates
+            instance variables.
+        """
+        
+        self.startTime = datetime.datetime.now().strftime("%Y-%m-%d-%H:%M")
+        
+        # Create directory for logs if it does not already exists
+        try:
+            os.mkdir("Logs")
```
yzhvictor, line 21: Instead of creating customized log mechanism, why not use python's logging library to log messages?

tools/unsupervised_dataset/Crawler/CrawlerLog.py
```diff
@@ -0,0 +1,71 @@
...
+        except FileExistsError as e:
+            pass
+        
+        self.log = open("Logs/log-" + self.startTime + ".txt", "a")
+        self.log.write("Beginning crawl at time {0}.\n".format(self.startTime))
+        
+        try:
+            os.mkdir("Queries")
+        except FileExistsError as e:
+            pass
```
yzhvictor, line 31: I think we need to handle this case more carefully. Imagine people mistakenly run the tool with an old path, it may unexpectedly overwrite some existing crawled data?

tools/unsupervised_dataset/Crawler/CQNode.py
```diff
@@ -0,0 +1,32 @@
+class CQNode:
```
yzhvictor, line 1: In general, it's good to have tests covering the actual logic. So maybe you can consider adding some corresponding tests for each source files.

tools/unsupervised_dataset/Crawler/CrawlerLog.py
```diff
@@ -0,0 +1,71 @@
...
+            instance variables.
+        """
+        
+        self.startTime = datetime.datetime.now().strftime("%Y-%m-%d-%H:%M")
+        
+        # Create directory for logs if it does not already exists
+        try:
+            os.mkdir("Logs")
+        except FileExistsError as e:
+            pass
```
kikkyo, line 23: I am not expert in Python, is this recommended practice?
I think we usually check for existence before making directories, instead of ignoring exceptions?

tools/unsupervised_dataset/Crawler/CrawlerLog.py
```diff
@@ -0,0 +1,71 @@
...
+        
+        self.log = open("Logs/log-" + self.startTime + ".txt", "a")
+        self.log.write("Beginning crawl at time {0}.\n".format(self.startTime))
+        
+        try:
+            os.mkdir("Queries")
+        except FileExistsError as e:
```
kikkyo, line 30: same here

tools/unsupervised_dataset/Crawler/run_crawler.py
```diff
@@ -0,0 +1,16 @@
+import requests
+import sys
+import Crawler
+
+# Initializes a crawler and starts the crawling process using command line arguments
+def start_crawler():
+    # TO-DO: Allow user to optionally define max size or max depth
```
kikkyo, line 7: nit: seems style guide recommends us using: # TODO(kl@gmail.com): Use a "*" here for string repetition.

tools/unsupervised_dataset/Crawler/README.md
```diff
@@ -0,0 +1,11 @@
+# SQL Crawler
+
+This directory contains the code to run a universal, unsupervised SQL web crawler. The user provides a starting target URL from which to begin crawling, and has the option to set the maximum depth or size of the crawler.
+
+## Usage
+To run the crawler, run the following command:
+
+```
+python3 run_crawler.py <starting URLs>
```
kikkyo, line 9: nit: add a comment on "URL should be  separated by comma or space/something else" ?

tools/unsupervised_dataset/Crawler/CQNode.py
kikkyo: +1 Please add tests0 
Assume pylint has been run on this?
Also do we already have some format checker integrated, it is easier to have people's code more consistent.

tools/unsupervised_dataset/SQLCrawler/CrawlerLog.py
```diff
@@ -0,0 +1,76 @@
...
+import csv
+import os
+import logging
+import pathlib
+
+class CrawlerLog(object):
+    """ Logs the status of the SQL crawler, including websites and queries.
+        
+        The CrawlerLog keeps track of which websites were explored, how many
+        queries were found, and creates a CSV with all the queries. It also
```
kikkyo, line 11: just a second thought on the output file format (not necessarily to be resolved in the cl:)), if we make it as CSV, later would it be problematic to read the query for later processing? For example, if a multi-line query is not well formatted.

PR #72, opened 2020-06-24
SQL Extraction: refactor to use sequences

creativitRy:
I did some refactoring on the SQL Extraction project

- fixed minor mistake in readme (run -> build)
- made relevant classes input/output sequences (like Java streams but in Kotlin) instead of some other data structure
- made DataFlowSolver retrieve the queries from the engine rather than the frontend
- renamed FileListExpander to FilesExpander
- made QueryFragment immutable
- implemented Comparable for Location

### DIFF
--- a/tools/sql_extraction/src/main/kotlin/DataFlowEngine.kt
+++ b/tools/sql_extraction/src/main/kotlin/DataFlowEngine.kt
@@ -20,7 +20,12 @@ class DataFlowEngine {
             .map { QueryUsages(it.key, it.value.sorted()) }
     }
 
-    private fun registerUsage(usage: Location, query: QueryFragment) {
+    /**
+     * Marks [usage] as one of the part of code where [query] is used.
+     *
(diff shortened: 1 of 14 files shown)


tools/sql_extraction/src/main/kotlin/DataFlowEngine.kt
```diff
@@ -1,7 +1,26 @@
...
+    /**
+     * Returns all query fragments and usages found by the engine thus far
+     */
+    fun getAllQueries(): Sequence<QueryUsages> {
+        return queryUsages.asSequence()
+            .filter { it.value.isNotEmpty() }
+            .map { QueryUsages(it.key, it.value.sorted()) }
+    }
+
+    private fun registerUsage(usage: Location, query: QueryFragment) {
```
feiling, line 23: Pls add comment.

I prefer to use name "addUsage". Also, it feels more natural to me to put the parameter "query" before "usage".

PR #73, opened 2020-06-25
[Automatic Query Fixer] init the project and implement a query token service

mingen-pan:
init a project with basic structures

create and test the QueryTokenService

PR #74, opened 2020-06-25
Auto Query Fixer: Implement QueryPositionConverter

mingen-pan:
Implement a QueryPositionConverter class, which can convert the line and col numbers to the index of a string. Test cases are included.

This tool will be used when the auto fixer tries to fix the syntax error. 

This PR is developed on #73


mingen-pan:
@kikkyo 
@Bei95

PR #75, opened 2020-06-26
Query Verification: Setup project and input handling

krishsuchdev:
Project setup using Maven and initial input handling for data free query verification.


tools/query_verification/src/main/java/com/google/bigquery/BigQueryInstance.java
```diff
@@ -0,0 +1,22 @@
+package com.google.bigquery;
+
+public class BigQueryInstance {
```
yzhvictor, line 3: It may be interesting to consider Autovalue: https://www.baeldung.com/introduction-to-autovalue

yzhvictor, line 3: Why this is "instance"?

tools/query_verification/src/main/java/com/google/bigquery/Main.java
```diff
@@ -0,0 +1,175 @@
...
+     *                              Referenced files should be in a JSON format.
+     *  -h,--help                   Print this help screen.
+     *
+     * @param args Command line arguments
+     */
+    public static void main(String[] args) {
+        Main main = new Main();
+        main.run(args);
+
+        System.exit(0);
```
yzhvictor, line 32: I don't think you need this.

tools/query_verification/src/main/java/com/google/bigquery/QVQuery.java
```diff
@@ -0,0 +1,21 @@
+package com.google.bigquery;
+
+public class QVQuery {
```
yzhvictor, line 3: Just put the full name, QV may be confusing for people, similar for other places. This can also consider Autovalue.

tools/query_verification/src/main/java/com/google/bigquery/BigQueryInstance.java
Luminarys, line 3: Please add some JavaDoc describing the class and its functionality

tools/query_verification/src/main/java/com/google/bigquery/QueryVerification.java
```diff
@@ -0,0 +1,85 @@
+package com.google.bigquery;
+
+public class QueryVerification {
+
+    private QVQuery migratedQuery;
```
yzhvictor, line 5: Should them be final and only set in constructor?

tools/query_verification/src/main/java/com/google/bigquery/QueryVerification.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public QueryVerification() {
+    }
+
+    public QueryVerification(QVQuery migratedQuery, QVSchema migratedSchema) {
+        this.migratedQuery = migratedQuery;
+        this.migratedSchema = migratedSchema;
+    }
+
+    public QueryVerification(QVQuery migratedQuery, QVSchema migratedSchema, QVQuery originalQuery, QVSchema originalSchema) {
```
yzhvictor, line 19: Will it be better to have a single constructor and mark the input params as @Nuallable

tools/query_verification/src/main/java/com/google/bigquery/QueryVerification.java
```diff
@@ -0,0 +1,85 @@
...
+        if (useDataAwareVerification) {
+            verificationResult = dataAwareVerification();
+        } else {
+            verificationResult = dataFreeVerification();
+        }
+
+        System.out.printf("Data-%s Verification %s\n", useDataAwareVerification ? "Aware" : "Free", verificationResult ? "Succeeded" : "Failed");
+    }
+
+    public boolean dataFreeVerification() {
```
yzhvictor, line 75: I can image these methods can be big, consider moving to their own class?

tools/query_verification/src/main/java/com/google/bigquery/QVQuery.java
Luminarys, line 3: Rather than give these a QV prefix can we change this to be a custom package called verification or query.verification?

Luminarys, line 3: Please add JavaDoc

tools/query_verification/src/main/java/com/google/bigquery/QVSchema.java
```diff
@@ -0,0 +1,20 @@
+package com.google.bigquery;
```
Luminarys, line 1: Same comments here as QVQuery

tools/query_verification/src/main/java/com/google/bigquery/QueryVerification.java
```diff
@@ -0,0 +1,85 @@
+package com.google.bigquery;
```
Luminarys, line 1: Please add some JavaDoc

tools/query_verification/src/main/java/com/google/bigquery/QueryVerification.java
```diff
@@ -0,0 +1,85 @@
...
+        boolean useDataAwareVerification = originalQuery != null && originalSchema != null;
+        boolean verificationResult;
+
+        if (useDataAwareVerification) {
+            verificationResult = dataAwareVerification();
+        } else {
+            verificationResult = dataFreeVerification();
+        }
+
+        System.out.printf("Data-%s Verification %s\n", useDataAwareVerification ? "Aware" : "Free", verificationResult ? "Succeeded" : "Failed");
```
Luminarys, line 72: Given you're using two ternaries here I think you could just add this line into the verification methods and have them return void

tools/query_verification/src/main/java/com/google/bigquery/QueryVerification.java
Luminarys, line 75: This is named like a getter, please change

tools/query_verification/src/main/java/com/google/bigquery/QueryVerification.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public boolean dataFreeVerification() {
+        // TODO Implement data free verification
+        return false;
+    }
+
+    public boolean dataAwareVerification() {
```
Luminarys, line 80: Same here

tools/query_verification/src/main/java/com/google/bigquery/Main.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    public void run(String[] args) {
+        CommandLine command = buildCommand(args);
+        if (command == null) {
+            return;
+        }
+
+        QueryVerification queryVerification = new QueryVerification();
```
Luminarys, line 41: I think it might look cleaner if you restructure things so that original/migratedQuery and schema get assigned to some variables at this scope, then create the QueryVerification object using the complete constructor. Then you can remove the setters as well.

tools/query_verification/src/main/java/com/google/bigquery/QueryVerification.java
```diff
@@ -0,0 +1,85 @@
+package com.google.bigquery;
+
+public class QueryVerification {
```
Luminarys, line 3: I think we should prefer to call this QueryVerifier

tools/query_verification/src/main/java/com/google/bigquery/Main.java
```diff
@@ -0,0 +1,175 @@
...
+        return options;
+    }
+
+    /**
+     * Retrieves the text contents from file
+     *
+     * @param path The path to the file to be read
+     * @return Text contents in file
+     */
+    public String getContentsOfFile(String path) {
```
Luminarys, line 163: We can make this a static method

PR #76, opened 2020-06-29
SQL Crawler: Implemented generic extraction

noah-kuo:
- Allows user to extract queries from a page, even without knowing HTML formatting


tools/unsupervised_dataset/sql_crawler/extraction_modules/generic_extraction_module.py
```diff
@@ -1,8 +1,32 @@
...
+        Returns
+            A list of queries in the form of strings.
+        """
+        # Remove HTML tags and special characters
+        content = html_response.text
+        converted = html.unescape(content)
+        tags_removed = re.sub('<[^<]+?>', '', converted)
+
+        # Look for text that matches common SQL queries
+        matches = re.findall(r"(?:SELECT|WITH|CREATE)[^;]*;", tags_removed)
```
yzhvictor, line 26: Maybe define this as a separate const string. I think this can be longer in the future when you cover insert, update, merge, etc.

tools/unsupervised_dataset/sql_crawler/extraction_modules/generic_extraction_module.py
```diff
@@ -1,8 +1,32 @@
...
+
+class GenericExtractionModule():
     """ A module to extract SQL queries from any website without knowing the HTML
         format of a site.
     """
 
-    def find_queries(html):
-        #TODO(Noah): Implement generic module
-        return []
+    def find_queries(html_response):
```
yzhvictor, line 10: Let's always have some test coverage for new code added, thanks.

tools/unsupervised_dataset/sql_crawler/extractor.py
```diff
@@ -64,5 +64,4 @@ def retrieve_module(url):
     if "cloud.google.com" in url:
```
kikkyo, line 64: Use const string too?

tools/unsupervised_dataset/sql_crawler/extraction_modules/generic_extraction_module.py
kikkyo: +1

PR #77, opened 2020-06-30
SQL Extraction: test and build with GitHub Actions workflow

creativitRy:
Implemented CI with GitHub Actions, existing independently alongside the existing Cloud Build project.

Once triggered, the workflow tests and builds the code using Java 1.8, and it uploads the build (containing the compiled code as a jar, the dependencies, and a script to run the jar) as an artifact (available on the actions tab above). It will only be triggered by changes to the project files on either a push or a pull request for any branch.

The badge added to the README file might not resolve properly until the workflow is merged in.


.github/workflows/sql-extraction.yml
```diff
@@ -0,0 +1,42 @@
+name: SQL Extraction
+
+on:
+  push:
```
kikkyo, line 4: If we don't filter on branches, it will be default run all branches right?

creativitRy: Yup. Should I restrict this to master?

feiling: I don't think this matters. Your call.

PR #80, opened 2020-07-01
Template generation—Initial PR

spoiledhua:
## Initial Commits

This is the [Template-Based Query Generation](https://docs.google.com/document/d/1QKxSlMs9TPzzWPadMDZb-1uRK2dtn9IGpD8FGOaIaFA/edit?usp=sharing) team's first PR. We've added keyword handler classes to parse user config files, create initial mappings between a user-friendly language and PostgreSQL/BigQuery, and constructed the Markov Chain class we'll be using to generate the random queries. We've also written initial tests for these classes.

## Upcoming Revisions/Known Issues

The parsers read from `.txt` files, which we intend to change in the next PR to a more structured file type, such as `.json` or `.xaml`.

The `UtilsTest` class doesn't have unit tests for functions that write to files. Are there any standard ways of unit testing write functions without actually creating new files?

The parsers use relative file paths to read from the `user_config` and `dialect_config` files. This was discussed with Bonan, who confirmed this approach is okay, but if there are better ways to achieve this let us know!

Currently `QueryGenerator` parses from the config files; we intend to change this such that `QueryGenerator` uses an instance of the `KeyWord` class to get the set of user-specified keywords instead, separating parsing and query generation between the classes.

The current keyword sets and mappings aren't exhaustive; we plan to complete these in the coming weeks.

## Next Commits

We'll focus on the revisions mentioned above, as well as create `Skeleton` classes for formatting generated queries.


tools/template_based_query_generation/src/main/java/QueryType.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public enum QueryType {
+    QUERY_ROOT,
+    DML,
+    DQL,
+    DDL,
+    SINK,
+    DDL_CREATE,
+    DDL_PARTITION,
+    DDL_CLUSTER,
```
Bei95, line 12: DML/DQL/DDL does not seem to be in the equivalent type as DDL_CREATE/DDL_PARTITION/... Consider splitting?

tools/template_based_query_generation/src/main/java/Utils.java
```diff
@@ -0,0 +1,185 @@
...
+		}
+
+		StringBuilder sb = new StringBuilder();
+
+		for (int i = 0; i < length; i++) {
+			int randomIndex = (int) (random.nextDouble() * CHARSET.length());
+			char randomChar = CHARSET.charAt(randomIndex);
+			if (i == 0 && Character.isDigit(randomChar)) {
+				// SQL identifiers can't start with digits, so replace with an arbitrary character
+				randomChar = 'a';
```
Bei95, line 56: consider use CHARSET.charAt(random.nextInt(0,52)) ?

tools/template_based_query_generation/src/main/java/Utils.java
```diff
@@ -0,0 +1,185 @@
...
+	public static String getRandomString(int length) throws IllegalArgumentException {
+		if (length <= 0) {
+			throw new IllegalArgumentException("Random string must have positive length");
+		}
+
+		StringBuilder sb = new StringBuilder();
+
+		for (int i = 0; i < length; i++) {
+			int randomIndex = (int) (random.nextDouble() * CHARSET.length());
+			char randomChar = CHARSET.charAt(randomIndex);
```
Bei95, line 53: CHARSET.charAt(random.nextInt(0,63))

tools/template_based_query_generation/src/test/java/UtilsTest.java
```diff
@@ -0,0 +1,53 @@
...
+		outputs.put("Postgre_skeletons", ImmutableList.copyOf(postgre_skeletons));
+		outputs.put("Postgre_tokenized", ImmutableList.copyOf(postgre_tokenized));
+		try {
+			Utils.writeDirectory(ImmutableMap.copyOf(outputs));
+		} catch (IOException exception) {
+			System.out.println(exception);
+		}
+		 */
+	}
+}
```
Bei95, line 53: Please add tests for this method and remove unused code.

tools/template_based_query_generation/src/test/java/QueryGeneratorTest.java
```diff
@@ -0,0 +1,27 @@
...
+                "dialect_config/root_dependencies.txt"};
+        String[] userConfigPaths = new String[] {"user_config/ddl.txt",
+                "user_config/dml.txt",
+                "user_config/dql.txt",
+                "user_config/root.txt"};
+        String mainUserConfig = "user_config/config.txt";
+        QueryGenerator qg = new QueryGenerator(dialectConfigPaths, userConfigPaths, mainUserConfig);
+        ArrayList<ArrayList<Query>> queries = qg.generateQueries(50);
+        for (ArrayList<Query> query: queries) {
+            System.out.println(query);
```
Bei95, line 23: Please use assert instead of print in test. Also, consider use test data in case real data is too long.

tools/template_based_query_generation/src/main/java/Node.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+        this.neighbors = edges;
+        this.updateProbabilities();
+    }
+
+    public ArrayList<Node<E>> getNeighborList() {
+        return neighborList;
+    }
+
+    public ArrayList<Double> getCProbabilities() {
```
Bei95, line 120: Consider to use getCumulativeProbabilities instead of getCProbabilities to make it more readable.

tools/template_based_query_generation/src/main/java/Utils.java
```diff
@@ -0,0 +1,185 @@
...
+
+		if (!file.exists() && !file.mkdir()) {
+			throw new FileNotFoundException("No such directory or the directory could not be created");
+		}
+
+		writeFile(outputs.get("BQ_skeletons"), outputDirectory + "/bq_skeleton.txt");
+		writeFile(outputs.get("BQ_tokenized"), outputDirectory + "/bq_tokenized.txt");
+		writeFile(outputs.get("Postgre_skeletons"), outputDirectory + "/postgre_skeleton.txt");
+		writeFile(outputs.get("Postgre_tokenized"), outputDirectory + "/postgre_tokenized.txt");
+		// TODO: write sample data to file
```
Bei95, line 83: Consider have assignment for each todo? For example "TODO(spoiledhua):..."

tools/template_based_query_generation/src/main/java/Keywords.java
```diff
@@ -0,0 +1,23 @@
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * Keywords set interface that all keyword set classes implement
+ */
+
+public interface Keywords {
```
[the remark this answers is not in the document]
spoiledhua: I believe we were expecting that these classes would allow Allen and I to work separately on DDL and DML respectively, but we've divided the work in a different way so yes I think it makes sense to just use the Keywords class and specify file_path. Will make that change soon.

tools/template_based_query_generation/src/main/java/KeywordsMapping.java
```diff
@@ -0,0 +1,22 @@
+/**
+ * Keywords mapping interface that all keyword mapping classes implement
+ */
+
+public interface KeywordsMapping {
```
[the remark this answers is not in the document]
spoiledhua: Similar reasoning to the Keywords class, will make the change soon.

tools/template_based_query_generation/src/main/java/Utils.java
spoiledhua: Good catch, will change it.

tools/template_based_query_generation/src/test/java/UtilsTest.java
spoiledhua: Okay, will add new tests using the Mock objects we discussed.

tools/template_based_query_generation/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <java.version>1.8</java.version>
+        <maven.compiler.version>3.6.0</maven.compiler.version>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
+    </properties>
+
+    <groupId>org.example</groupId>
```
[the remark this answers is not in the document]
AllenWang314: Changed to  <groupId>com.google.bigquery</groupId>

tools/template_based_query_generation/src/main/java/MarkovChain.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.HashMap;
+import java.util.Random;
+
+/**
+ * Class that represents src.main.MarkovChain encoding query and keyword dependencies
+ */
+public class MarkovChain<E> {
+    
+    private HashSet<Node<E>> nodes;
+    private Random r;
```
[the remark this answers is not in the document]
AllenWang314: Yes, originally I had the randomness happen in MarkovChain, but I later moved it to Node, so it's no longer needed here. I have removed it and refactored code from QueryGenerator.

tools/template_based_query_generation/src/main/java/MarkovChain.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * constructs MarkovChain object from edge weights and random seed
+     * @param nodes
+     * @param seed
+     */
+    public MarkovChain(HashMap<Node<E>, HashMap<Node<E>, Double>> nodes, int seed) {
```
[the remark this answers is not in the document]
AllenWang314: The config will have a seed field that will be by default set to 0, the user can choose not to change it in the config.

tools/template_based_query_generation/src/main/java/Node.java
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * class representing node for a query in the markov chain
+ */
+public class Node<E> {
+
+    private E obj;
+    private HashMap<Node<E>, Double> neighbors;
+    private ArrayList<Node<E>> neighborList; // list of neighbors and corresponding cumulative probabilities
+    private ArrayList<Double> cProbabilities; // cumulative probabilities
+    private Random r;
```
bonancao, line 16: Should we keep a Random instance in each node or all of the nodes should share the same instance?

tools/template_based_query_generation/src/main/java/Utils.java
```diff
@@ -50,10 +51,10 @@ public static String getRandomString(int length) throws IllegalArgumentException
 
 		for (int i = 0; i < length; i++) {
 			int randomIndex = (int) (random.nextDouble() * CHARSET.length());
```
Bei95, line 53: Removed unused variable: randomIndex

tools/template_based_query_generation/src/main/java/Utils.java
```diff
@@ -50,10 +51,10 @@ public static String getRandomString(int length) throws IllegalArgumentException
...
-			char randomChar = CHARSET.charAt(randomIndex);
+			char randomChar = CHARSET.charAt(random.nextInt(0, 63));
```
Bei95, line 54: Sorry for asking for change back and forth. I guess it would be better to use CHARSET.charAt(random.nextInt(0, CHARSET.length())) here.

tools/template_based_query_generation/src/main/java/MarkovChain.java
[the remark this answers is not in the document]
Bei95: If we have a default seed value in config, does that mean if user does not notice that, we will always use the same seed and generate the same output?

tools/template_based_query_generation/src/main/java/Node.java
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * class representing node for a query in the markov chain
+ */
+public class Node<E> {
+
+    private E obj;
+    private HashMap<Node<E>, Double> neighbors;
+    private ArrayList<Node<E>> neighborList; // list of neighbors and corresponding cumulative probabilities
+    private ArrayList<Double> cProbabilities; // cumulative probabilities
```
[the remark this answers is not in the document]
AllenWang314: I've changed it to Treemap storing both cProbabilities and neighborList. Now that I think about it, there might not be a point in having HashMap<Node<E>, Double> neighbors as a field.

tools/template_based_query_generation/src/main/java/Node.java
AllenWang314: All of them share the same instance, since they are initialized in the same class with the seed given by the user. Is it a problem if they share the same instance? Would they be generating similar random numbers then? Is it a problem if they have different instances?

tools/template_based_query_generation/src/main/java/Node.java
```diff
@@ -0,0 +1,124 @@
...
+        this.r = new Random(seed);
+        this.setNeighbors(new HashMap<Node<E>, Double>());
+    }
+
+    /**
+     * updates neighborList and cProbabilities when neighbors is changed
+     */
+    private void updateProbabilities() {
+        if (this.neighbors.size() != 0) {
+            Set<Node<E>> neighborSet = this.neighbors.keySet();
```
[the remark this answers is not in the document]
AllenWang314: fixed!

tools/template_based_query_generation/src/main/java/Node.java
AllenWang314: fixed!

tools/template_based_query_generation/src/main/java/QueryType.java
AllenWang314: I see what you mean: two types are needed but the tricky part is MarkovChain and Node need objects of the same type. What I plan is to have to still keep MarkovChain<Query> and Node<Query> but I will change the structure of Query in a later PR (adding stuff about number of tokens needed, etc)

tools/template_based_query_generation/src/main/java/Node.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.TreeMap;
+import java.util.Random;
+
+/**
+ * class representing node for a query in the markov chain
+ */
+public class Node<E> {
+
+    private E obj;
+    private HashMap<Node<E>, Double> neighbors;
```
Bei95, line 12: We still store duplicated data in this neighbors hashmap and the following cumulativeProbabilities. Even though the storage space is not sensitive in this project, storing same info in different places and also expect those 2 objects storing identical nodes could possible cause more errors in the future when we or someone else edit it. Do we really need to store this neighbor variable? Looks like the probability was only used to calculated the cumulativeProbability. If the probability need to be kept for some other use, can we store it in the cumulativeProbabilities object? For example, use TreeMap<Double, Pair<Node<E>, Double>>

tools/template_based_query_generation/src/main/java/QueryType.java
Bei95: Please add a todo here.

tools/template_based_query_generation/src/main/java/QueryGenerator.java
```diff
@@ -0,0 +1,142 @@
...
+     * @param mainUserConfig
+     * @throws Exception
+     */
+    public QueryGenerator(String[] dialectConfigPaths, String[] userConfigPaths, String mainUserConfig) throws Exception {
+        // read in lines from userConfigPaths, ignoring lines that begin with ' ' or '/'
+        // stores active queries in activatedQueries
+        HashSet<String> activatedQueries = new HashSet<String>();
+        String line;
+        for (String path: userConfigPaths) {
+            InputStream inputStream = ClassLoaderUtil.getResourceAsStream(path, QueryGenerator.class);
```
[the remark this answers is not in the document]
Bei95: As discussed offline, this file will be changed after we moved to gson. Please add a todo in this file as well as the test file with some short explanation of the proposing change. Thank you!

tools/template_based_query_generation/src/main/java/Node.java
Bei95: With the current implementation I believe it each not will have its own instance as the instance is initialized in the contructor?
Also, vote +1 for sharing the same instance. If they do not share the same instance, I **guess** with the same seed each node will get same list of random numbers.

tools/template_based_query_generation/src/main/java/Keywords.java
```diff
@@ -0,0 +1,23 @@
...
+
+	/**
```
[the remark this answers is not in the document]
AllenWang314: I think this is fixed now---Victor revised this file significantly

tools/template_based_query_generation/src/main/java/Node.java
AllenWang314: I think neighbors would come in handy if we ever needed to modify the markov chain. However, from how we imagine people using our tool, the markov chain will be initialized and used to generate queries and won't be modified. The user will modify the config and recreate a markov chain to make modifications. So I think it's ok to completely get rid of neighbors.

tools/template_based_query_generation/src/main/java/KeywordsMappingDDL.java
```diff
@@ -0,0 +1,55 @@
...
+
+	/**
+	 * Constructor of DDL keywords mapping, parsed from the config file
+	 */
+	public KeywordsMappingDDL() {
+		keywordsMapPostgre = Utils.makeImmutableMap(fileNamePostgre, keywordsSetDDL);
+
+		keywordsMapBQ = Utils.makeImmutableMap(fileNameBQ, keywordsSetDDL);
+
+		System.out.println(System.getProperty("user.dir"));
```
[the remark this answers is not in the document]
spoiledhua: Done!

tools/template_based_query_generation/src/main/java/Node.java
AllenWang314: I kept neighbors because it was nice having it for testing (though it's never used by the other classes). I got rid of it and made edits to NodeTest to only use cumulativeProbabilities.

tools/template_based_query_generation/src/main/java/MarkovChain.java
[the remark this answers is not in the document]
AllenWang314: Bei pointed me to Random() which solved this problem!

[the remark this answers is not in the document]
AllenWang314: No randomness is needed in Markov Chain so I removed this constructor, see below for further comments on Nodes sharing the same Random r.

tools/template_based_query_generation/src/main/java/Node.java
AllenWang314: When the Nodes are created in QueryGenerator, they will be given the same Random r. I have added a constructor in Node to take in an instance of Random.

tools/template_based_query_generation/src/main/java/QueryGenerator.java
```diff
@@ -0,0 +1,142 @@
...
+    private Node<Query> queryRoot;
+
+    /**
+     *
+     * @param dialectConfigPaths
+     * @param userConfigPaths
+     * @param mainUserConfig
+     * @throws Exception
+     */
+    public QueryGenerator(String[] dialectConfigPaths, String[] userConfigPaths, String mainUserConfig) throws Exception {
```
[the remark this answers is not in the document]
AllenWang314: I was mainly using QueryGenerator as a prototype to print some strings out and see what it does. Victor is working on replacing this completely using gson, and the body of this class has been removed and replaced with a TODO

tools/template_based_query_generation/src/test/java/QueryGeneratorTest.java
AllenWang314: The body of QueryGenerator and QueryGeneratorTest have been removed. Victor is working on doing it via a json approach,.

tools/template_based_query_generation/src/main/java/Utils.java
AllenWang314: Victor made a fix.

tools/template_based_query_generation/src/main/java/Utils.java
AllenWang314: Changed to char randomChar = CHARSET.charAt(random.nextInt(0, CHARSET.length()));

tools/template_based_query_generation/src/main/java/Utils.java
AllenWang314: Victor made this change.

tools/template_based_query_generation/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+            <!--                            <goal>add-source</goal>-->
+            <!--                        </goals>-->
+            <!--                        <configuration>-->
+            <!--                            <sources>-->
+            <!--                                <source>source-path-goes-here</source>-->
+            <!--                            </sources>-->
+            <!--                        </configuration>-->
+            <!--                    </execution>-->
+            <!--                </executions>-->
+            <!--            </plugin>-->
```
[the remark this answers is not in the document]
AllenWang314: Done

tools/template_based_query_generation/src/main/java/QueryType.java
AllenWang314: Yup added!

tools/template_based_query_generation/src/main/resources/user_config/dql.json
```diff
@@ -0,0 +1,49 @@
+{
+<<<<<<< HEAD
```
Bei95, line 2: Please resolve those merging conflicts

tools/template_based_query_generation/src/test/java/QueryGeneratorTest.java
```diff
@@ -0,0 +1,36 @@
+import org.junit.jupiter.api.Test;
+
+<<<<<<< HEAD
```
Bei95, line 3: Please resolve those lined caused by merging conflicts.

tools/template_based_query_generation/src/main/resources/user_config/dql.json
AllenWang314: Ahh, sorry I missed the test folder and config files. Just patched this part.

tools/template_based_query_generation/src/test/java/QueryGeneratorTest.java
AllenWang314: Fixed

tools/template_based_query_generation/src/main/java/Utils.java
AllenWang314: Yup assigned to Victor

tools/template_based_query_generation/src/main/java/QueryGenerator.java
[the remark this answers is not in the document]
AllenWang314: Added TODOs assigned to Allen

tools/template_based_query_generation/src/main/java/QueryGenerator.java
```diff
@@ -0,0 +1,142 @@
...
+        // parse lines from mainUserConfig, ignoring lines that begin with ' ' or '/'
+        InputStream inputStream = ClassLoaderUtil.getResourceAsStream(mainUserConfig, QueryGenerator.class);
+        InputStreamReader streamReader = new InputStreamReader(inputStream, StandardCharsets.UTF_8);
+        BufferedReader br = new BufferedReader(streamReader);
+        while ((line = br.readLine()) != null) {
+            if (line.length() > 2 && line.charAt(0) != ' '
+                    && !(line.charAt(0) == '/' && line.charAt(1) == '/')) {
+                String[] res = line.split(":");
+                if(res[0].equals("root")) {
+                    this.queryRoot = nodes.get(res[1]);
```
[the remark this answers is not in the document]
AllenWang314: Entire file has been changed.

tools/template_based_query_generation/src/main/java/Keywords.java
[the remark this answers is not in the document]
spoiledhua: This should be fixed now.

tools/template_based_query_generation/src/main/java/Utils.java
spoiledhua: Fixed!

tools/template_based_query_generation/src/main/java/Utils.java
spoiledhua: Fixed.

tools/template_based_query_generation/src/main/java/MarkovChain.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Random;
+
+/**
+ * Class that represents src.main.MarkovChain encoding query and keyword dependencies
+ */
+public class MarkovChain<E> {
+
+	private HashSet<Node<E>> nodes;
+
+	/**å
```
Bei95, line 14: Please delete the weird character

AllenWang314: got it!

PR #81, opened 2020-07-01
SQL Extraction: dataflow engine variable scope and environment

creativitRy:
Environment is the data structure that the dataflow engine uses to keep track of all possible query fragments a variable can have.


tools/sql_extraction/src/main/kotlin/DataFlowEngine.kt
```diff
@@ -20,6 +24,42 @@ class DataFlowEngine {
             .map { QueryUsages(it.key, it.value.sorted()) }
     }
 
+    /**
+     * Visits method scope from an Antlr visitor
```
kikkyo, line 28: nit: Visits method scope from an Antlr visitor**.** (with periods)

tools/sql_extraction/src/main/kotlin/DataFlowEngine.kt
```diff
@@ -20,6 +24,42 @@ class DataFlowEngine {
...
+     */
+    fun visitMethod(visitChildren: () -> Unit) {
+        visitScope(visitChildren)
+    }
+
+    /**
+     * Associates [paramName] as a method argument for the most recently called [visitMethod]
+     */
+    fun addMethodParameter(paramName: String) {
+        LOGGER.debug("add method parameter $paramName")
```
kikkyo, line 40: nit: This comment is intentionally leave there for debugging purpose? 
If just for logging, should we use LOGGER.log (if there is a log method :)).. and we should use "Add method parameter $paramName."

tools/sql_extraction/src/main/kotlin/DataFlowEngine.kt
```diff
@@ -3,12 +3,16 @@ package com.google.cloud.sqlecosystem.sqlextraction
...
+private val LOGGER = KotlinLogging.logger { }
 
 /**
  * Backend engine for running data-flow analysis
+ *
  * @see FrontEnd
  */
-class DataFlowEngine {
+class DataFlowEngine(private val environment: Environment = Environment()) {
     private val queryUsages: HashMap<QueryFragment, HashSet<Location>> = HashMap()
```
kikkyo, line 16: nit: in java we usually just use Map<QueryFragment, HashSet<Location>> = HashMap() if it is the same in kotlin..

tools/sql_extraction/src/main/kotlin/DataFlowEngine.kt
```diff
@@ -3,12 +3,16 @@ package com.google.cloud.sqlecosystem.sqlextraction
 import com.google.cloud.sqlecosystem.sqlextraction.output.Location
 import com.google.cloud.sqlecosystem.sqlextraction.output.QueryFragment
 import com.google.cloud.sqlecosystem.sqlextraction.output.QueryUsages
+import mu.KotlinLogging
+
+private val LOGGER = KotlinLogging.logger { }
```
kikkyo, line 8: Is this a global logger? should this be inside the class?

tools/sql_extraction/src/main/kotlin/Environment.kt
```diff
@@ -0,0 +1,96 @@
...
+ * A mapping from variable name to all possible query fragment values
+ */
+class Environment {
+    private var parentScope: Environment?
+    private var reachingDefs: HashMap<String, QueryFragment?>
+
+    constructor() {
+        this.parentScope = null
+        this.reachingDefs = HashMap()
+    }
```
kikkyo, line 15: Should we use init block instead?

tools/sql_extraction/src/main/kotlin/DataFlowEngine.kt
creativitRy: This was suggested by the [framework](https://github.com/MicroUtils/kotlin-logging/wiki#obtaining-a-logger) as the preferred way of obtaining a logger. Otherwise, it increases boilerplate code.

tools/sql_extraction/src/main/kotlin/DataFlowEngine.kt
creativitRy: Map is immutable in Kotlin, but I can change it to MutableMap

tools/sql_extraction/src/main/kotlin/Environment.kt
creativitRy: That would execute for the other compiler as well, which is redundant. It might look cleaner though.

tools/sql_extraction/src/main/kotlin/Environment.kt
```diff
@@ -0,0 +1,96 @@
+package com.google.cloud.sqlecosystem.sqlextraction
+
+import com.google.cloud.sqlecosystem.sqlextraction.output.QueryFragment
+
+/**
+ * A mapping from variable name to all possible query fragment values
+ */
+class Environment {
+    private var parentScope: Environment?
+    private var reachingDefs: HashMap<String, QueryFragment?>
```
feiling, line 10: The name `reachingDefs` is less than ideal. I would use something like `variableUsages` or `variableReferences`.

tools/sql_extraction/src/main/kotlin/Environment.kt
```diff
@@ -0,0 +1,96 @@
...
+    fun declareVariable(varName: String) {
+        reachingDefs[varName] = null
+    }
+
+    /**
+     * Checks whether a variable of name [varName] was declared in this current scope.
+     *
+     * @return true if variable was declared in this scope
+     */
+    fun hasVariableInScope(varName: String): Boolean {
```
feiling, line 35: I prefer name `isVariableDeclaredInScope`.

tools/sql_extraction/src/main/kotlin/Environment.kt
```diff
@@ -0,0 +1,96 @@
...
+        return varName in reachingDefs
+    }
+
+    /**
+     * Gets all possible queries for the variable [varName].
+     * Variable can be declared in any reachable scope.
+     *
+     * @throws[NullPointerException] if variable does not exist
+     */
+    fun getVariable(varName: String): QueryFragment? {
```
feiling, line 45: I prefer name like `getVariableUsage` or `getVariableReference`

tools/sql_extraction/src/main/kotlin/Environment.kt
```diff
@@ -0,0 +1,96 @@
...
+            parentScope!!.getVariable(varName)
+        }
+    }
+
+    /**
+     * Gets all possible queries for the variable [varName].
+     * Variable can be declared in any reachable scope.
+     * [default] is returned if variable doesn't exist in any reachable scope.
+     */
+    fun getVariableOrDefault(varName: String, default: QueryFragment? = null): QueryFragment? {
```
feiling, line 58: Similarly, `getVariableUsageOrDefault` or `getVariableReferenceOrDefault`.

tools/sql_extraction/src/main/kotlin/Environment.kt
```diff
@@ -0,0 +1,96 @@
...
+        return when {
+            hasVariableInScope(varName) -> reachingDefs[varName]
+            parentScope != null -> parentScope!!.getVariableOrDefault(varName)
```
feiling, line 61: Seems `parentScope!!.getVariableOrDefault(varName)` should be `parentScope!!.getVariableOrDefault(varName, default)`?

tools/sql_extraction/src/main/kotlin/Environment.kt
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    /**
+     * Overwrites the possible queries for the variable [varName] existing in the most recent scope.
+     * Variable needs to be defined first to be set.
+     *
+     * @throws[NullPointerException] if variable does not exist
+     */
+    fun setVariable(varName: String, query: QueryFragment?) {
```
feiling, line 72: Will `setVariable` be called multiple times for the same variable?

tools/sql_extraction/src/main/kotlin/Environment.kt
creativitRy: Oh oops, thanks for pointing that out!

tools/sql_extraction/src/main/kotlin/Environment.kt
creativitRy: Yes, it will be recursively called until the correct scope is found.

tools/sql_extraction/src/main/kotlin/Environment.kt
creativitRy: Update: using init would break any code using the constructor without arguments.

PR #86, opened 2020-07-07
Query Fixer: reorganize the previous PR and init the project

mingen-pan:

PR #87, opened 2020-07-08
Auto Query Fixer: Implement BigQueryService

mingen-pan:
BigQueryService class is used to communicate with the BigQuery server. It is primarily used to dry run a query.

This PR is after #86


tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryParserFactory.java
```diff
@@ -11,12 +11,13 @@
...
 
 import com.google.cloud.bigquery.utils.queryfixer.exception.ParserCreationException;
+import com.google.common.base.Preconditions;
 
 import java.io.Reader;
-import java.util.Objects;
 
 /**
- * A factory to generate parsers. The fault generated parser is Babel Parser with BigQuery dialect.
+ * A factory to generate parsers. The default generated parser is Babel Parser with BigQuery
```
kikkyo: nit: is '**a**' Babel Paser

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/QueryTokenProcessorTest.java
```diff
@@ -10,21 +11,20 @@
 
 public class QueryTokenProcessorTest {
 
-  private QueryTokenProcessor createService() {
-    return new QueryTokenProcessor(new BigQueryParserFactory());
+  private QueryTokenProcessor tokenService;
+
+  @Before
```
kikkyo: Since we only need this once, you can use @BeforeClass?

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryParserFactory.java
mingen-pan: Done.

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/QueryTokenProcessorTest.java
mingen-pan: Good point

mingen-pan: I think `@Before` is better because I need it to init a member variable. `@BeforeClass` is a static method and cannot init a member variable.

PR #88, opened 2020-07-08
Query Breakdown: Initialize Project

subinbean:
Wrote a sketch of what the project would look like: contains roughly all the classes that I think are necessary and how they will interact (added some methods/fields). A simple test of the parser logic is added to ensure that testing works. The project builds correctly using maven--the dependency file is included as well.


tools/query_breakdown/src/main/java/Node.java
```diff
@@ -0,0 +1,33 @@
+import java.util.ArrayList;
+
+/**
+ * This class represents a single node in the tree visualization of the algorithm. Below are fields
+ * necessary for implementation.
+ */
+public class Node {
+
+  // ensures tree structure
+  ArrayList<Node> children;
```
feiling, line 10: Please use explicit access control, i.e. add "private"  for this member. Also do the same for other members.

PR #89, opened 2020-07-08
Query Verification: Data free verification with json schema

krishsuchdev:
Implemented data free verification for schema provided as a JSON file.

### DIFF
--- a/tools/query_verification/src/main/java/com/google/bigquery/BigQuerySchemaJsonDeserializer.java
+++ b/tools/query_verification/src/main/java/com/google/bigquery/BigQuerySchemaJsonDeserializer.java
@@ -20,17 +20,19 @@ public class BigQuerySchemaJsonDeserializer implements JsonDeserializer<FieldLis
         for (JsonElement fieldElement : fieldsArray) {
             JsonObject fieldObject = fieldElement.getAsJsonObject();
 
-            // Field class uses LegacySQLTypeName for type
+            // Field class stores types as LegacySQLTypeName instead of StandardSQLTypeName
             if (fieldObject.has("type")) {
+                // Convert standard type to legacy type
                 StandardSQLTypeName standardType = StandardSQLTypeName.valueOf(fieldObject.get("type").getAsString());
                 LegacySQLTypeName legacyType = LegacySQLTypeName.legacySQLTypeName(standardType);
 
+                // Insert LegacySQLTypeName object so it can be used for type
                 JsonObject typeObject = new JsonObject();
                 typeObject.addProperty("constant", legacyType.name());
                 fieldObject.add("type", typeObject);
             }
 
-            // Field class uses subFields instead of fields
+            // Field class uses subFields instead of fields for STRUCT/RECORD type
             if (fieldObject.has("fields")) {
                 fieldObject.add("subFields", fieldObject.get("fields"));
                 fieldObject.remove("fields");
--- a/tools/query_verification/src/main/java/com/google/bigquery/BigQuerySchemaJsonDeserializer.java
+++ b/tools/query_verification/src/main/java/com/google/bigquery/BigQuerySchemaJsonDeserializer.java
@@ -1,0 +1,45 @@
+package com.google.bigquery;
+
+import com.google.cloud.bigquery.Field;
+import com.google.cloud.bigquery.FieldList;
+import com.google.cloud.bigquery.LegacySQLTypeName;
+import com.google.cloud.bigquery.StandardSQLTypeName;
+import com.google.gson.*;
+
+import java.lang.reflect.Type;
+
+/**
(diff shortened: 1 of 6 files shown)


tools/query_verification/src/test/resources/schema1.json
```diff
@@ -0,0 +1,11 @@
+[
+  {
+    "schema": {
```
yzhvictor, line 3: Do you need this if it only contains 1 property, will this work:

[
  {
    "fields": [],
     "tableReference": {
           ....
      }
  }
]

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -54,4 +101,28 @@ public void verifyDataAware() {
...
     }
 
+    /**
+     * Reads JSON schema to create table fields based on the schema
+     * @param queryVerificationSchema Schema to read from
+     * @return New table info
+     */
+    public static TableInfo getTableInfoFromJsonSchema(QueryVerificationSchema queryVerificationSchema) {
+        // TODO Support multiple table schema
+        JsonObject schemaObject = queryVerificationSchema.getJsonArray().get(0).getAsJsonObject();
```
yzhvictor, line 111: Should we check length before accessing?

tools/query_verification/src/main/java/com/google/bigquery/BigQuerySchemaJsonDeserializer.java
```diff
@@ -0,0 +1,45 @@
...
+public class BigQuerySchemaJsonDeserializer implements JsonDeserializer<FieldList> {
+
+    @Override
+    public FieldList deserialize(JsonElement jsonElement, Type type, JsonDeserializationContext jsonDeserializationContext) throws JsonParseException {
+        JsonArray fieldsArray = jsonElement.getAsJsonArray();
+
+        for (JsonElement fieldElement : fieldsArray) {
+            JsonObject fieldObject = fieldElement.getAsJsonObject();
+
+            // Field class uses LegacySQLTypeName for type
```
yzhvictor, line 23: Maybe add more comments regarding why using LegacySQLTypeName

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -38,7 +44,48 @@ public void verify() {
...
+        if (migratedSchema != null) {
+            try {
+                if (migratedSchema.isInJsonFormat()) {
+                    TableInfo tableInfo = QueryVerifier.getTableInfoFromJsonSchema(migratedSchema);
+                    Table table = bigQuery.create(tableInfo);
+                    tables.add(table);
+                } else {
+                    // TODO Load schema from DDL
+                }
+            } catch (NullPointerException e) {
```
Luminarys, line 61: Although this saves some lines of code I'd advise against catching a NPE as it could be triggered for a variety of reasons. If possible I'd recommend handling null objects in the verifier code even if it makes a bit more verbose.

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -38,7 +44,48 @@ public void verify() {
...
+        JobInfo jobInfo = JobInfo.newBuilder(queryConfig)
+                .setJobId(jobId)
+                .build();
+
+        // TODO Support multiple queries
+
+        try {
+            // Run dry-run
+            Job queryJob = bigQuery.create(jobInfo);
+            verificationResult = queryJob.getStatistics() != null;
```
Luminarys, line 80: I assume in the case of a bad query we'd get back some error information from queryJob, it's probably worth collecting that if we can.

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -38,7 +44,49 @@ public void verify() {
...
+                .build();
+
+        // TODO Support multiple queries
+
+        try {
+            // Run dry-run
+            Job queryJob = bigQuery.create(jobInfo);
+            verificationResult = queryJob.getStatus().getState() == JobStatus.State.DONE;
+        } catch (BigQueryException e) {
+            // Print out syntax/semantic errors returned from BQ
```
Luminarys, line 82: Have you tested an exception is thrown due to a bad query, not just due to something like a failure to connect/bad credentials etc?

krishsuchdev: I've found that if the query is invalid from a syntax or semantic error, BQ throws an exception so I can just print the error message in catch block.

Luminarys: Ok that should be good for now then, thanks!

PR #90, opened 2020-07-09
SQL Crawler: Support for command-line arguments and more SQL expressions, additional test cases

noah-kuo:


tools/unsupervised_dataset/run_crawler.py
```diff
@@ -1,11 +1,16 @@
...
 
 import sys
+import argparse
 from sql_crawler import crawler
 
 def start_crawler():
-    urls = sys.argv[1:]
-    new_crawler = crawler.Crawler(urls, max_size=50)
+    parser = argparse.ArgumentParser(description="SQL Web Crawler")
+    parser.add_argument("urls", help="A list of URLs to be crawled", nargs='+')
```
yzhvictor, line 9: Is the list separated by comma or something? maybe add that as well in description

tools/unsupervised_dataset/run_crawler.py
```diff
@@ -1,11 +1,16 @@
...
+    parser.add_argument("--max_depth", help="The max depth of the crawler (default=3)", type=int, nargs=1, default=[3])
+    parser.add_argument("--max_size", help="The maximum number of links to be crawled (default=100)", type=int, nargs=1, default=[100])
+    args = parser.parse_args()
+    new_crawler = crawler.Crawler(args.urls, max_size=args.max_size[0], max_depth=args.max_depth[0])
```
yzhvictor, line 13: Is this the standard? looks like an array, but it should be an int I think.

tools/unsupervised_dataset/sql_crawler/crawler_log.py
```diff
@@ -66,11 +71,68 @@ def log_error(self, errorMessage):
...
+    def set_gcs(self, location):
+        """ Sets variables for uploading data to Google Cloud Storage.
+            
+        Args:
+            location: String with name of project ID and bucket name,
+            separated by a period.
+        """
+
+        self.gcs_project, self.gcs_bucket = self.parse_location_arg(location)
+        if self.gcs_project:
```
yzhvictor, line 103: Do we also need to check gcs_bucket, and bq_dataset for the below method?

PR #91, opened 2020-07-10
Query Verification: Support for multiple table schema and ddl

krishsuchdev:
Implemented data free verification for schema provided in DDL and support for schema files consisting of multiple tables.


tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -51,26 +51,38 @@ public void verifyDataFree() {
...
+                // Schema is DDL
+                JobInfo jobInfo = configureJob(migratedSchema.schema(), false);
+                Job schemaJob = bigQuery.create(jobInfo);
+                try {
+                    schemaJob.waitFor();
+                } catch (InterruptedException e) {
+                    System.out.println(e.getMessage());
+                }
+
+                List<TableId> tableIds = QueryVerifier.getTableIdsFromDdlSchema(migratedSchema);
```
Luminarys, line 72: Does the schema job not return information about the table created that we can use here instead of reparsing the mgiratedSchema?

krishsuchdev: I wasn't able to find that. I think the only table information it returns are the query results, which is empty in this case.

Luminarys: Testing with the BQ UI it seems like [this api](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults) is used which returns the table schema. Can you see if it's possible to use this?

krishsuchdev: It looks like `getQueryResults` just returns null for ddl, so it doesnt return a schema for me.

Luminarys: Interesting. If it doesn't work then I guess this is fine. Thanks for checking!

PR #92, opened 2020-07-10
SQL Extraction: Java analyzer

creativitRy:
The Java Visitor is generated by Antlr and involves lots of boilerplate code. It visits all nodes in the parsed AST in program order, and each node gets a callback where they can specify which child nodes are visited: this is either done by automatically visiting them all (`visitChildren`) or by calling `accept(this)` manually for each child node. The methods that are not overriden visit all children by default.

The idea is to decorate some visit callbacks with the relevant dataflow engine method. These include finding a literal, concatenating strings, managing variable scope, etc. The dataflow engine assumes the frontend is visiting nodes correctly, and it keeps string literals in reference for both variables and the current expression. The current expression is reset to null after each statement.


tools/sql_extraction/src/test/kotlin/JavaAnalysisTest.kt
```diff
@@ -0,0 +1,151 @@
+package com.google.cloud.sqlecosystem.sqlextraction
+
+import com.google.cloud.sqlecosystem.sqlextraction.output.QueryUsages
+import org.antlr.v4.runtime.CharStreams
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class JavaAnalysisTest {
+    @Test
+    fun `strings without usages are not returned`() {
```
creativitRy, line 10: Maybe this test should belong to DataFlowEngine rather than JavaFrontEnd.

PR #93, opened 2020-07-10
SQL Extraction: Basic query confidence rater

creativitRy:
Before I implement a SQL library/framework-specific analyzer, I wrote a way of rating confidence for basic detected fragments. This simply involves counting how many SQL keywords are present in each fragment. Keywords are defined to have weighted values, so some keywords can be given higher/lower weight (via code) if needed. If the first word in the fragment is a keyword, then it is given extra weight.
Finally, the summed weights are converted to a value between 0 and 1 skewed to the left.


tools/sql_extraction/src/main/kotlin/ConfidenceRater.kt
```diff
@@ -0,0 +1,113 @@
...
+import com.google.cloud.sqlecosystem.sqlextraction.output.QueryFragment
+
+private const val WITH_DELIMITER = "((?<=%1\$s)|(?=%1\$s))"
+
+/**
+ * Rates a [QueryFragment] a confidence value to indicate how likely it is to be a query.
+ * @see rate
+ */
+class ConfidenceRater {
+    private val keywords: Map<String, Int> = mapOf(
```
feiling, line 12: This should be a Set, instead of Map.

tools/sql_extraction/src/main/kotlin/ConfidenceRater.kt
```diff
@@ -0,0 +1,113 @@
...
+        if (query.complex != null) {
+            return query.complex.sumBy(this::countSqlLikeTokens)
+        } else if (query.literal.isNullOrBlank()) {
+            return 0
+        }
+
+        var count = 0
+        for ((index, token) in tokenize(query.literal).withIndex()) {
+            var localCount = keywords.getOrDefault(token.toUpperCase(), 0)
+            if (index == 0 && localCount > 0) {
```
feiling, line 95: What's the purpose of checking `index==0`?

creativitRy: This was to check if a keyword appears at the front of the string, which I assumed would happen more frequently if the string is an actual SQL query.

feiling: Please add a comment to explain this.

PR #100, opened 2020-07-15
Auto Query Fixer: add a new method to the BigQueryService component.

mingen-pan:
Add the `listTables` method to the BigQueryService.
Test is included.


tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
```diff
@@ -0,0 +1,89 @@
...
+   * */
+  public BigQueryService(BigQueryOptions options) {
+    this.bigquery = new BigQueryOptions.DefaultBigQueryFactory().create(options);
+  }
+
+  /**
+   * Initialize a connection to BigQuery server with the default setting. The projectID should be specified.
+   * @param projectId project ID
+   * */
+  public BigQueryService(String projectId) {
```
kikkyo, line 37: check not null?

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
```diff
@@ -0,0 +1,89 @@
...
+    BigQueryOptions.Builder builder = BigQueryOptions.newBuilder()
+        .setProjectId(projectId);
```
kikkyo, line 39: nit: You can just  .build() here.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
```diff
@@ -0,0 +1,89 @@
...
+   * @return the job representing this dry run
+   * @throws BigQueryException the error from the BigQuery Server.
+   */
+  public Job dryRun(String query) throws BigQueryException {
+    QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).setDryRun(true).build();
+    return bigquery.create(JobInfo.of(queryConfig));
+  }
+
+  /**
+   * return the BigQuery Exception if the dry run of a query generated the exception. If the query is correct and no
```
kikkyo, line 55: nit: **R**eturn

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
```diff
@@ -0,0 +1,89 @@
...
+   * errors exist in the server, null will be returned.
+   * @param query the dry-run query
+   * @return the BigQueryException related with this query
+   */
+  public BigQueryException catchExceptionFromDryRun(String query) {
```
kikkyo, line 60: I think normally we don't indicate if exception is caught or thrown  "catException.." in the method name. This method can be called "dryRun" - the other method can be merged together here. 
I don't quite understand why you want to catch and return? You can just throw it.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
```diff
@@ -0,0 +1,89 @@
...
+      dryRun(query);
+    } catch (BigQueryException exception) {
+      return exception;
+    }
+
+    return null;
+  }
+
+  /**
+   * Fetch the names of all table from the BigQuery given the project and dataset. If the project and/or dataset do
```
kikkyo, line 71: nit: The line break looks weird. I think you should not hit "new line" manually when typing otherwise after reformatting it would look weird :).

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
```diff
@@ -0,0 +1,89 @@
...
+   * not exist or not visible to the provided service account, {@link com.google.cloud.bigquery.BigQueryException} will
+   * be thrown.
+   * @param projectId project id
+   * @param datasetId dataset id
+   * @return list of table names belonging to the given project and dataset
+   */
+  public List<String> listTableNames(String projectId, String datasetId) throws BigQueryException {
+    DatasetId projectDatasetId = DatasetId.of(projectId, datasetId);
+    Page<Table> tables = bigquery.listTables(projectDatasetId);
```
kikkyo, line 80: How many tables can be returned at maximum?  If there are 10k+  tables, how long does it take roughly?

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.collection.IsIterableContainingInOrder.contains;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class BigQueryServiceTest {
+
+  private BigQueryService service;
+
+  /**
```
kikkyo, line 24: What does it mean here? Where do we read in the credentials?

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+    return builder.build();
+  }
+
+  @Before
+  public void getService() {
+    service = new BigQueryService(getOptions());
+  }
+
+  @Test
+  public void dryRun1() {
```
kikkyo, line 42: dryRun_success

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+    assertEquals("corpus", fields.get(0).getName());
+  }
+
+  @Test
+  public void dryRun2() {
+    String query = "SELECT corpus FROM `bigquery-public-data.samples.shakespearex`";
+    BigQueryException exception = service.catchExceptionFromDryRun(query);
+    assertNotNull(exception);
+    assertEquals("Not found: Table", exception.getMessage().substring(0, 16));
+  }
```
kikkyo, line 57: If you want to test different cases for the same flow, consider using parameterized test, we should avoid using dryRun1, dryRun2, dryRun3...

kikkyo: https://www.tutorialspoint.com/junit/junit_parameterized_test.htm

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
mingen-pan: Given the page structure, it can download as many as it can if we use call `nextPage`. Here to make it simple, I will add an constraint to allow this method to only fetch max 1000 tables. Given a Table only contains name information if not reloaded, so the traffic shouldn't be heavy.

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
mingen-pan: This method requires connection with BigQuery server. 

Uses can call `gcloud auth application-default login` and the credential will stored in the default path, which will be read by this method.

However, uses can download a service account credential and put it elsewhere. If this is the case, they need to modify the test cases manually.

I will put this demonstration in the comment

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
mingen-pan: Thanks for providing this.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
mingen-pan: I rerun it. It seems auto format will break every 100 char

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
mingen-pan: It is also possible to mock the BigQuery server, but I am not sure how long it will take. I guess connecting to real BigQuery and uses the public datasets may be an convenient way to test?

mingen-pan:
Mock BigQuery is added. The unit test will not connect to the BigQuery server

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
```diff
@@ -1,52 +1,69 @@
...
 
 /**
  * A service to connect with the BigQuery server. It is used to communicate with the server like
  * sending queries and receiving data.
- * */
+ */
 public class BigQueryService {
 
-  private final BigQuery bigquery;
+  private static final int TableFetchSize = 1000;
```
kikkyo, line 25: Its recommended to use UPPER_SNAKE_CASE for constant variable names. "TABLE_FETCH_SIZE"

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
```diff
@@ -59,4 +76,29 @@ public BigQueryException catchExceptionFromDryRun(String query) {
...
     return null;
   }
+
+  /**
+   * Fetch the names of all table from the BigQuery given the project and dataset. If the project
+   * and/or dataset do not exist or not visible to the provided service account, {@link
+   * com.google.cloud.bigquery.BigQueryException} will be thrown.
+   *
+   * <p>It is possible that a dataset contains a large amount of tables, so this method will only
+   * fetch maximum 1000 of them.
```
kikkyo, line 86: 1000 seems a bit small.. :) Maybe you can adjust this later.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
```diff
@@ -59,4 +76,29 @@ public BigQueryException catchExceptionFromDryRun(String query) {
...
+   *
+   * @param projectId project id
```
kikkyo, line 88: I don't the "param" comment make many sense so far(they are mostly self-explanary), as seen from other methods. Later you can just keep the comment of the method but leave out the comments for the params.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
```diff
@@ -59,4 +76,29 @@ public BigQueryException catchExceptionFromDryRun(String query) {
...
+   * @param datasetId dataset id
+   * @return list of table names belonging to the given project and dataset
+   */
+  public List<String> listTableNames(String projectId, String datasetId) throws BigQueryException {
+    DatasetId projectDatasetId = DatasetId.of(projectId, datasetId);
+    Page<Table> tables =
+        bigQuery.listTables(projectDatasetId, BigQuery.TableListOption.pageSize(TableFetchSize));
+    return StreamSupport.stream(tables.iterateAll().spliterator(), /* parallel= */ false)
```
kikkyo, line 96: what if you set the "parallel= true", can you add a comment why you decided to set it to false and when we have to set it to true?

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
```diff
@@ -42,11 +54,85 @@ public void dryRun1() {
...
+    assertEquals(2, tables.size());
+    assertThat(tables, contains("311_request", "311_service_requests"));
+  }
+
+  /**
+   * Mock the connection to the BigQuery server. Use reflection to inject the fake connection into
+   * the {@link BigQueryService}.
+   */
+  private void fakeBigQuery() {
+    BigQuery bigQuery = mock(BigQuery.class);
```
kikkyo, line 78: Since the service is going to be shared, we should init the variable before all tests starts and use mock annotation.

kikkyo: use "@Mock" instead of mock()

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
mingen-pan: Yes, I think I can make it as an optional argument later for users to specify.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
mingen-pan: Sounds good!

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
mingen-pan: Here may not be a good place to use `@mock`. BigQuery object will be injected into the `BigQueryService` field, which will be shared by different tests. The BigQuery object will be only mocked once in the beginning, because it is only called by the `@Before` method.

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
```diff
@@ -69,4 +69,70 @@ public void listTables() {
...
+   * Mock the response of dry run if "SELECT corpus FROM `bigquery-public-data.samples.shakespeare`
+   * GROUP BY corpus limit 1000" is given.
+   */
+  private void correctShakespeare(BigQuery bigQuery) {
+    String query =
+        "SELECT corpus FROM `bigquery-public-data.samples.shakespeare` GROUP BY corpus limit 1000";
+    QueryJobConfiguration queryConfig =
+        QueryJobConfiguration.newBuilder(query).setDryRun(true).build();
+
+    Job job = mock(Job.class);
```
kikkyo, line 98: Why you need to mock this simple object? you can create a default instance job instance

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
```diff
@@ -69,4 +69,70 @@ public void listTables() {
...
+    JobStatistics.QueryStatistics queryStatistics = mock(JobStatistics.QueryStatistics.class);
```
kikkyo, line 99: Same there, just use a default instance. We usually don't mock out these simple objects.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
mingen-pan: Comment is added

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/service/BigQueryService.java
mingen-pan: Done

mingen-pan:
Mock test is refactored.

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/BigQueryServiceTest.java
```diff
@@ -30,109 +31,105 @@
...
    */
-  private void fakeBigQuery() {
-    BigQuery bigQuery = mock(BigQuery.class);
-    correctShakespeare(bigQuery);
-    incorrectShakespeare(bigQuery);
-    fakeTables(bigQuery);
+  private void injectBigQueryMockIntoService() {
     try {
-      FieldUtils.writeField(service, "bigQuery", bigQuery, true);
+      FieldUtils.writeField(service, "bigQuery", bigQueryMock, true);
```
kikkyo, line 102: nit: /* blahcomment = true*/

PR #102, opened 2020-07-15
SQL Extraction: show progress flag, show debug logs flag

creativitRy:
1. Instead of controlling log levels through the properties file, I got it to emit debug logs by adding the `--debug` flag.
2. I added a flag `--progress` that prints out how much work is done to STDERR. This also means that the sequence of file paths to analyze is expanded first so that the total amount of needed work can be found.
3. Unknown file types are now ignored rather than throwing an exception.


tools/sql_extraction/src/main/kotlin/DataFlowSolver.kt
```diff
@@ -29,6 +29,7 @@ class DataFlowSolver(private val frontends: List<FrontEnd>) {
             }
         }
 
-        throw IllegalArgumentException("Cannot analyze $filePath. No frontend available.")
+        LOGGER.warn { "Cannot analyze $filePath. No frontend available. Skipping file..." }
```
kikkyo, line 32: nit: should you print the filePath somewhere? $filePath doesn't have much information or it would be replaced by the actual variable?

tools/sql_extraction/src/test/kotlin/DataFlowSolverTest.kt
```diff
@@ -39,13 +42,16 @@ class DataFlowSolverTest {
     }
 
     @Test
-    fun `solver throws without any valid frontends`() {
+    fun `solver returns empty sequence without any valid frontends`() {
         val frontEnd = mockk<FrontEnd>()
```
kikkyo, line 46: Is it better practice to use mock annotation @MockK?

kikkyo: I think it is more common in our codebase to use annotation and set up the 'shared' mock variable inside the test class, instead of creating them in each method.

tools/sql_extraction/src/test/kotlin/DataFlowSolverTest.kt
```diff
@@ -39,13 +42,16 @@ class DataFlowSolverTest {
     }
 
     @Test
-    fun `solver throws without any valid frontends`() {
+    fun `solver returns empty sequence without any valid frontends`() {
```
kikkyo, line 45: curious on the style too... why not 'SolverReturnsEmptySequence....'? Is this recommended?

tools/sql_extraction/src/main/kotlin/DataFlowSolver.kt
creativitRy: $ is string interpolation in Kotlin. It should be replaced by the actual variable's toString().

tools/sql_extraction/src/test/kotlin/DataFlowSolverTest.kt
creativitRy: Since the language allows method names with spaces, I thought it would be more readable. I just did more reading and found [unit testing best practices for Kotlin](https://resources.jetbrains.com/storage/products/kotlinconf2018/slides/4_Best%20Practices%20for%20Unit%20Testing%20in%20Kotlin.pdf), which is recommending backticks.

tools/sql_extraction/src/main/kotlin/DataFlowSolver.kt
kikkyo: thanks!

tools/sql_extraction/src/test/kotlin/DataFlowSolverTest.kt
kikkyo: ok thanks!

tools/sql_extraction/src/main/kotlin/Cli.kt
```diff
@@ -74,13 +72,28 @@ private class Cli(
         "--pretty", help = "Pretty-print output JSON"
     ).flag()
 
+    private val showProgress: Boolean by option(
+        "--progress", help = "Print progress to STDERR"
+    ).flag()
+
+    private val debug: Boolean by option(
+        "--debug", "--verbose", help = "Print debug logs to STDERR"
```
kikkyo, line 80: Do we have to specify both options to make it work?

tools/sql_extraction/src/main/kotlin/Cli.kt
```diff
@@ -74,13 +72,28 @@ private class Cli(
...
+    ).flag()
+
     override fun run() {
-        LOGGER.debug("Starting SQL Extraction from command line")
+        if (debug) {
+            System.setProperty(org.slf4j.impl.SimpleLogger.DEFAULT_LOG_LEVEL_KEY, "Debug")
```
kikkyo, line 85: This setProperty looks a bit hard coded, do we have to update this if our dependency/libs change in future?

tools/sql_extraction/src/main/kotlin/Cli.kt
creativitRy: It works for either one

tools/sql_extraction/src/main/kotlin/Cli.kt
creativitRy: Yes, this would have to change if we were to change the logging dependencies from slf4j to something else.

creativitRy: I realized I can import SimpleLogger for brevity.

PR #103, opened 2020-07-15
Query Breakdown: Deletion implemented, version 1 of input parsing complete, loop sketched

subinbean:
The deletion mechanism is implemented and tested for single and multiple line queries: by catching the exception, we can extract successfully the line/column of where the exception occurs. Using that information, we can manipulate the string to delete the token: instances such as group>3 where tokens are not separated by spaces is also dealt with. 

Other than that, input parsing works but needs to be updated by having new queries separate by new lines. The function for the algorithm is also sketched.


tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
```diff
@@ -1,13 +1,17 @@
...
  */
-public class QueryBreakdown {
+public class Main {
   public static void main(String[] args) {
     String inputFile = null;
-    double timeLimit = 0.0;
+    int errorLimit = 0;
     String outputFile = null;
 
     if (args[0] != null) {
```
feiling, line 17: The logic seems wrong here. Seems it should be 
```
if (args[0] == null) {
```

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
```diff
@@ -17,19 +21,24 @@ public static void main(String[] args) {
       System.out.println("You have not specified the input file");
```
feiling, line 21: When this happens, shouldn't you need to throw an exception to stop the execution?

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
```diff
@@ -17,19 +21,24 @@ public static void main(String[] args) {
...
     }
     if (args[2] != null) {
       outputFile = args[2];
     }
 
     // this is where we will put the file I/O logic through the input reader.
-    String originalQuery = InputReader.readInput(inputFile);
+    String originalQuery = null;
+    try {
+      originalQuery = InputReader.readInput(inputFile);
```
feiling, line 33: `inputFile` is not initialized.

tools/query_breakdown/src/main/java/com/google/bigquery/InputReader.java
```diff
@@ -0,0 +1,38 @@
...
+    StringBuilder sb = new StringBuilder();
+
+    while (currentLine != null) {
+      sb.append(currentLine);
+      currentLine = reader.readLine();
+    }
+
+    reader.close();
+
+    return sb.toString();
```
feiling, line 36: The code can be simplified by calling `readAllLines`.

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
```diff
@@ -17,19 +21,24 @@ public static void main(String[] args) {
       System.out.println("You have not specified the input file");
     }
     if (args[1] != null) {
-      timeLimit = Double.parseDouble(args[1]);
+      errorLimit = Integer.parseInt(args[1]);
     }
     if (args[2] != null) {
       outputFile = args[2];
     }
```
feiling, line 29: Please create a work item to use a library such as http://commons.apache.org/proper/commons-cli/ to process flags.

tools/query_breakdown/src/main/java/com/google/bigquery/Node.java
```diff
@@ -7,22 +9,23 @@
...
+  private int startLine;
+  private int startColumn;
+  private int endLine;
+  private int endColumn;
 
   // ensures that we know how the error was handled
-  boolean errorHandlingType;
-  String replaceFrom;
-  String replaceTo;
+  private boolean errorHandlingType;
```
feiling, line 23: Use an enum instead of boolean for `errorHandlingType`.

subinbean: Hopefully resolved with the most recent commit!

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
subinbean: Addressed in most recent commit :)

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
subinbean: I think the logic here was that if an argument was provided, the variable would be initialized. I maintained the same logic for my new code--let me know if this is wrong!

tools/query_breakdown/src/main/java/com/google/bigquery/CalciteParser.java
```diff
@@ -12,10 +14,10 @@
  * Calcite Parser can be used as a blackbox.
  */
 public class CalciteParser implements Parser {
-  SqlParser.Config config;
+  private SqlParser.Config config;
```
AayushBhatnagar, line 17: should we make it a final field?

tools/query_breakdown/src/main/java/com/google/bigquery/InputReader.java
```diff
@@ -0,0 +1,43 @@
...
+    for (String line : lines) {
+      sb.append(line);
+    }
+
+    String parsedInput = sb.toString();
+    int len = parsedInput.length();
+    boolean lastSemicolon = false;
+
+    // deals with case where last query ends with a semicolon
+    if (parsedInput.charAt(len - 1) == ';') {
```
AayushBhatnagar, line 31: Why not first replace and then substring? In this case you end up doing substring, replace and append.

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+    // termination condition: if the parsing doesn't throw exceptions, then the leaf is reached
+    return;
+  }
+
+  /**
+   * This method implements the deletion mechanism: given the position of the component, it
+   * generates a new query with that component deleted.
+   */
+  static String deletion(String inputQuery, int startLine, int startColumn,
```
AayushBhatnagar, line 70: private?

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+/**
+ * This class is where the main logic lives for the algorithm that this tool utilizes. It will
+ * also be in charge of outputting the results.
+ */
+public class QueryBreakdown {
+
+  // global fields that keeps track of the minimum unparseable component so far
+  private int minimumUnparseableComp = Integer.MAX_VALUE;
```
AayushBhatnagar, line 16: If all methods are static, then why these fields are not static?

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
```diff
@@ -0,0 +1,128 @@
...
+  public static void run(String originalQuery, String outputFile, int errorLimit) {
+
+    // determines which parser to use
+    parser = new CalciteParser();
+
+    // uses the loop function to generate and traverse the tree of possible error recoveries
+    loop(originalQuery, errorLimit);
+
+    // write termination logic for output (tracing the node back, reconstructing path, output)
+    return;
```
AayushBhatnagar, line 37: no need for this statement.

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * This is where the code for the algorithm will go: essentially, there will be a loop that
+   * constantly inputs a new query after adequate error handling
+   */
+  private static void loop(String inputQuery, int errorLimit) {
+    try {
+      parser.parseQuery(inputQuery);
+    } catch (Exception e) {
+      // generates new queries through deletion and replacement
+      SqlParserPos pos = ((SqlParseException) e).getPos();
```
AayushBhatnagar, line 49: is there a cast failure possibility?

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
```diff
@@ -0,0 +1,128 @@
...
+          ((SqlParseException) e).getExpectedTokenNames());
+
+      // recursively loops through the new queries
+      loop(deletionQuery, errorLimit);
+      for (String s: replacementQueries) {
+        loop(s, errorLimit);
+      }
+    }
+    // termination condition: if the parsing doesn't throw exceptions, then the leaf is reached
+    return;
```
AayushBhatnagar, line 63: no need for this statement.

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
```diff
@@ -0,0 +1,128 @@
...
+
+    // write termination logic for output (tracing the node back, reconstructing path, output)
+    return;
+  }
+
+  /**
+   * This is where the code for the algorithm will go: essentially, there will be a loop that
+   * constantly inputs a new query after adequate error handling
+   */
+  private static void loop(String inputQuery, int errorLimit) {
```
AayushBhatnagar, line 44: Whats the use of errorLimit variable?

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
feiling: in this case, the code block should be something like this:

```java
if (args[0] != null) {
  inputFile = args[0];
} else {
  System.out.println("You have not specified the input file");
  return -1;
}
```

In the current form, `inputFile` is not initialized.

tools/query_breakdown/src/main/java/com/google/bigquery/InputReader.java
```diff
@@ -0,0 +1,43 @@
...
+      parsedInput = parsedInput.substring(0, len - 1);
+      lastSemicolon = true;
+    }
+
+    parsedInput = parsedInput.replaceAll(";", ";\n");
+    if (lastSemicolon) {
+      parsedInput = parsedInput + ';';
+    }
+
```
feiling, line 40: Please add a //TODO comment that a more robust way to break the input into statements will be implemented in the future. The current code probably works for simple cases, but it doesn't handle comments, or semicolon in strings for example.

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
```diff
@@ -0,0 +1,95 @@
...
+
+import java.io.IOException;
+import java.util.Comparator;
+
+/**
+ * this file is the main file for the command line tool. The first argument is a txt file
+ * containing the queries. The second argument, which is optional, is a time limit that we
+ * impose on processing a single query. The third argument, which is optional, is a file
+ * that we want to write the results to.
+ */
```
feiling, line 13: Please add comment describes what command line flags are supported. Also, add some sample usages.

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
```diff
@@ -0,0 +1,95 @@
...
+   * This is the method that instantiates a CommandLine object for the Apache CLI Interface. It deals with
+   * command line parsing as well as help generation once parsing is unavailable
+   */
+  public static CommandLine createCommand(String[] args) {
+    CommandLineParser parser = new DefaultParser();
+    Options options = createOptions();
+    HelpFormatter help = new HelpFormatter();
+
+    help.setOptionComparator(new Comparator<Option>() {
+      public int compare(Option option1, Option option2) {
```
feiling, line 61: What's the purpose of having this comparator?

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
```diff
@@ -0,0 +1,95 @@
...
+
+    return cl;
+  }
+
+  /**
+   * This is the method that instantiates options for the Apache CLI interface
+   */
+  public static Options createOptions() {
+    Options options = new Options();
+    options.addOption(Option.builder("r").required(true).longOpt("read").hasArg(true).argName("PATH")
```
feiling, line 87: I think "i" and "inputFile" is a better name for this flag.

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
```diff
@@ -0,0 +1,95 @@
...
+            .desc("this is the file containing queries to be inputted into the tool").build());
+    options.addOption(Option.builder("w").longOpt("write").hasArg(true).argName("PATH")
```
feiling, line 89: Use "o" and "outputFile" for this flag.

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
subinbean: Gotcha, I changed the code in main.java to reflect the CLI interface, so I do believe the current way should resolve this issue. Resolving for now!

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
subinbean: The purpose of the comparator is to just order the commands in the order I want it, since the default is alphabetical. 

Before: -r -l -w
After: -r -w -l

tools/query_breakdown/src/main/java/com/google/bigquery/InputReader.java
subinbean: fixed accordingly!

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
subinbean: left as package private for now as sort of a "test hook." Will address privacy settings later.

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
subinbean: An issue is that the exception can catch other unwanted exceptions: design will be changed in the future by creating an exception superclass

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
AayushBhatnagar: I can still see this return statement. can you please remove it first?

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
AayushBhatnagar: can you please add a comment in the code explaining the justification of extended visibility of the function?

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
subinbean: yikes, my bad--should be removed now

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
subinbean: yup, it's more of a comment that's applicable for the entire class so just added a comment above the entire thing

PR #106, opened 2020-07-16
Query Verification: Support for multiple queries

krishsuchdev:
Added support for multiple query statements for data free verification.


tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -70,37 +66,49 @@ public void verifyDataFree() {
...
             if (tables.isEmpty()) {
                 System.out.println(migratedSchema.path() + " is not correctly formatted.");
             }
         }
 
-        // Create dry-run job
-        JobInfo jobInfo = configureJob(migratedQuery.query(), true);
+        // Create dry-run jobs
+        List<JobInfo> jobInfos = getJobInfosFromQuery(migratedQuery, true);
+        ListIterator<JobInfo> jobInfoListIterator = jobInfos.listIterator();
```
Luminarys, line 79: I think we should prefer a loop like ```for (int i = 0; i < jobInfos.size(); i++) { JobInfo jobInfo = jobInfos.get(i); ...```  over using listIterator, I believe it should be more clear as opposed to using the iterator methods

tools/query_verification/src/test/java/com/google/bigquery/QueryVerifierTest.java
```diff
@@ -102,4 +103,34 @@ public void testGetMultipleTableIdFromDdlSchema() {
...
+    @Test
+    public void testGetMultipleJobInfosFromQuery() {
+        String queryContents = "SELECT * FROM table1;\nSELECT column1 FROM table2; SELECT column2 FROM table2;";
+        QueryVerificationQuery query = QueryVerificationQuery.create(queryContents, "");
+        List<JobInfo> jobInfos = QueryVerifier.getJobInfosFromQuery(query, true);
+
+        assertEquals(jobInfos.size(), 3);
+
+        List<String> queries = jobInfos.stream().map(jobInfo -> {
+            QueryJobConfiguration queryJobConfiguration = jobInfo.getConfiguration();
```
Luminarys, line 127: I would inline this into the return and then remove the {} from the lambda so it looks like ```jobInfo -> jobInfo.getConfiguration().getQuery()``` (I think this is valid syntax)

krishsuchdev: Only reason I didn't do that is because that actually requires manual casting since `.getConfiguration()` returns a `JobConfiguration` type (super class of `QueryJobConfiguration`) and doesn't have a `.getQuery()` method

Luminarys: ```QueryJobConfiguration queryJobConfiguration = jobInfo.getConfiguration();``` doesn't seem to have a downcast, is it not needed when you assign like this?

krishsuchdev: So it looks like `JobConfiguration` is an abstract class and `.getConfiguration()` instead returns `<C extends JobConfiguration>`

PR #107, opened 2020-07-17
SQL Extraction: refactor tests to use mock annotation

creativitRy:
Fix for https://github.com/GoogleCloudPlatform/bigquery-utils/pull/102#discussion_r455468113


feiling:
Looks good to me. @kikkyo, could you take a look also? I think you suggested this change to Gahwon.

PR #109, opened 2020-07-17
Allen wang314/adding unit tests

AllenWang314:
The number of files changed was getting too big, so making a pull request halfway through to receive some feedback.

- Restructured project into packages
- Added DataType and corresponding config file
- Added Table class to represent a data table schema
- Added Token and Tokenizer, restructured to include Token Info
- Added TokenInfo and Token to Query class
- Added additional "fail" tests to MarkovChainTest
- Created and implemented TableTest, make other test files with TODOs for a later PR


tools/template_based_query_generation/src/main/java/Query.java
```diff
@@ -1,16 +1,51 @@
...
 
-	private QueryType type;
+  private QueryType type;
+  private Mapping mapping;
+  private boolean special;
+  private Token token;
+
+  public Query(QueryType type, boolean special) {
+      this.type = type;
+      this.special = special;
```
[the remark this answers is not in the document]
AllenWang314: Got it

tools/template_based_query_generation/src/main/java/FeatureType.java
```diff
@@ -0,0 +1,28 @@
+/**
+ * Types of queries in SQL language encoding
+ * feature_root, ddl_feature_root, dml_feature_root, dql_feature_root, feature_sink are types to help create reference nodes
+ */
+public enum FeatureType {
+	feature_root, // type for the root node
```
Bei95, line 6: Sorry for the confusion! Let's still keep those types upper case

AllenWang314: got it!

PR #111, opened 2020-07-21
Template-Based Query Generation: Skeleton Class

spoiledhua:
Initial version of the Skeleton class. Assuming an input and output consisting of a List of Strings, the Skeleton class converts from our third language to both PostgreSQL and BigQuery, as well as appends necessary tokens after each keyword.

Once Allen finishes his changes to the Token and Query classes, I will refactor Skeleton to assume input and output of a List of Query objects instead.


tools/template_based_query_generation/src/main/java/parser/Utils.java
```diff
@@ -214,7 +159,7 @@ private static String getOutputDirectory(String directoryName) {
 
     ImmutableMap.Builder<String, ImmutableList<Mapping>> builder = ImmutableMap.builder();
 
-    for (Feature feature : features.getFeatures()) {
+    for (Feature feature : ImmutableList.copyOf(features.getFeatures())) {
```
bonancao, line 162: You probably won't need to make a immutable copy of the list here if you are only using it here and you are not modifying the list.

spoiledhua: Changed, and will look for similar occurrences throughout.

PR #113, opened 2020-07-21
Query Verification: Populate tables in bq and get query results

krishsuchdev:
Implemented the first part of the BQ side of data aware verification.


tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -113,11 +93,122 @@ public void verifyDataFree() {
...
+                writer.write(ByteBuffer.wrap(queryVerificationData.contents().getBytes()));
+                writer.close();
+            } catch (IOException e) {
+                System.out.println("I/O Exception: " + e.getMessage());
+            }
+
+            // Run table data writing job
+            try {
+                writer.getJob().waitFor();
+            } catch (InterruptedException e) {
```
Luminarys, line 208: Can we get more context from this than just the fact it was interrupted

tools/query_verification/src/main/java/com/google/bigquery/Main.java
```diff
@@ -82,9 +89,22 @@ public static void main(String[] args) {
...
         if (command.hasOption("d")) {
-            // TODO Data input for data aware verification
+            String[] dataOptionValues = command.getOptionValues("d");
+
+            for (String dataFilePath : dataOptionValues) {
+                String dataFileName = new File(dataFilePath).getName();
+                String dataContents = getContentsOfFile(dataFilePath);
+
+                // Check file format
+                if (dataFileName.substring(dataFileName.lastIndexOf(".") + 1).toLowerCase().equals("csv")) {
```
yzhvictor, line 99: Do we really want to check the extension. I feel sometimes although the files are csv format, they don't necessary have .csv extension. I think maybe we can just assume people are providing valid files?

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -113,11 +93,122 @@ public void verifyDataFree() {
...
+
+            QueryJobResults results = null;
+            try {
+                // Run query job
+                Job queryJob = bigQuery.create(jobInfo);
+
+                results = QueryJobResults.create(query, queryJob.getQueryResults());
+            } catch (BigQueryException | InterruptedException e) {
+                // Print out syntax/semantic errors returned from BQ
+                System.out.printf("Error in Query #%d from %s\n%s\n\n", i, migratedQuery.path(), e.getMessage());
```
yzhvictor, line 120: consider using logger library instead of system.out

yzhvictor: As discussed, we'll have a follow up pr to handle logging.

PR #117, opened 2020-07-25
SQL Extraction: highlight queries in VS Code

creativitRy:
Highlights all query fragments and query usages. Different queries are highlighted in different colors.
image

Also un-gitignored the necessary .vscode folder that was autogenerated.

### DIFF
--- a/tools/vscode_sql_extraction/.gitignore
+++ b/tools/vscode_sql_extraction/.gitignore
@@ -3,3 +3,4 @@ node_modules
 .vscode-test/
 *.vsix
 /resources/sql_extraction
+!/.vscode/
--- a/tools/vscode_sql_extraction/src/highlighter.ts
+++ b/tools/vscode_sql_extraction/src/highlighter.ts
@@ -11,6 +11,7 @@ export class Highlighter {
 
   /**
    * Highlights all queries in the open document.
+   *
    * @param openEditor currently open editor.
    * @param workspaceRoot root path of the open workspace.
    * @param queries list of found queries.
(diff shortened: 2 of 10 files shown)


tools/vscode_sql_extraction/.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules
 .vscode-test/
 *.vsix
 /resources/sql_extraction
+!/.vscode/
```
feiling, line 6: Are all those files under the .vscode directory needed to build the plugin?

tools/vscode_sql_extraction/src/highlighter.ts
```diff
@@ -0,0 +1,100 @@
...
+      return filePath === openPath;
+    });
+  }
+
+  /**
+   * Updates the cache if needed, and then returns the relevant decoration.
+   *
+   * @param index decoration number to return.
+   */
+  private getColor(index: number): vscode.TextEditorDecorationType {
```
feiling, line 88: In another pull request, please change getColor() to use the highlight color of the current theme.

tools/vscode_sql_extraction/.gitignore
creativitRy: I believe so. The samples on GitHub all have those 4 files.

tools/vscode_sql_extraction/src/highlighter.ts
creativitRy: Good idea! Currently, I'm highlighting each query using a different color, but I can just have a few theme colors and cycle between them.

PR #119, opened 2020-07-27
Query Breakdown: Program now executes successfully with deletion (w/o location tracking)

subinbean:
### **TLDR**
- Files to look at: QueryBreakdown, Node
- The program now executes with deletion and outputs the correct result! (which means that the run, loop, deletion function are all working properly). The caveat is that the location in which the error occurs is not tracked correctly yet (an issue discovered late last week) but will be fixed in the next next PR. 
- The PR is big in terms of code changed because I was initially going to make 1 big PR but realized (and learned) that PR's should be smaller. As such, for this PR I'm asking for a review on just QueryBreakdown and Node files pertaining to the program's execution for deletion! There will be code in QueryBreakdown pertaining to replacement but I commented it out as it will be the next PR. 

### **Specifics**
- The Node class has changed to better reflect the information that it should contain. 
- QueryBreakdown has some major changes explained below:
1. It is refactored as an object-based class: details are in the comments
2. The main loop as well as the run command is implemented and works seamlessly with deletion
3. Deletion is refactored to be more concise using the returnIndex function

### **Future To-Do's/PR's**
- Next PR: the main loop works correctly with replacement (w/o location tracking)
- Next, Next PR: the program now supports correct location tracking 
- Ahead after that: redesign the exception class to wrap SqlParseException (which should fix casting errors), format output into a JSON format or text file, support runtime limits


tools/query_breakdown/src/main/java/com/google/bigquery/Node.java
```diff
@@ -27,15 +27,91 @@
...
+    this.replaceTo = replaceTo;
+    this.unparseableCount = unparseableCount;
+  }
+
+  /**
+   * Override toString method for better debugging
+   */
+  @Override
+  public String toString() {
+    return "Parent: " + parent.getStartColumn() + parent.getEndColumn()
```
AayushBhatnagar, line 72: Nit: try to use str.format over + from the readability standpoint.

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
```diff
@@ -16,52 +15,109 @@
...
-   * it to the specified output file, or if that is null, generate a new file to put the output in.
-   * The provided timeLimit will stop the tool from running over a certain time.
+   * Constructor for the QueryBreakdown object. We model this class as an object rather than
+   * through static methods because the user should be able to call QueryBreakdown multiple
+   * times and create multiple instances of it.
    */
-  public static void run(String originalQuery, String outputFile, int errorLimit) {
+  public QueryBreakdown(Parser parser) {
+    minimumUnparseableComp = Integer.MAX_VALUE;
+    root = new Node();
```
AayushBhatnagar, line 32: inconsistent var naming. either use this. for all members or none at all.

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
```diff
@@ -16,52 +15,109 @@
...
+   *
+   * TODO: implement errorLimit logic, deal with exception casting
    */
-  private static void loop(String inputQuery, int errorLimit) {
+  private void loop(String inputQuery, int errorLimit, Node parent, int depth) {
     try {
       parser.parseQuery(inputQuery);
     } catch (Exception e) {
       // generates new queries through deletion and replacement
       SqlParserPos pos = ((SqlParseException) e).getPos();
```
AayushBhatnagar, line 87: Please verify that this a sqlparseexception type before casting it into one something like

if (e instanceof SqlParseException) {
....
}

better still catch SqlParseException besides catching superclass Exception.

tools/query_breakdown/src/main/java/com/google/bigquery/ReplacementLogic.java
```diff
@@ -1,7 +1,7 @@
 package com.google.bigquery;
 
 import java.util.ArrayList;
-import java.util.List;
+import java.util.HashSet;
```
AayushBhatnagar, line 4: remove this import.

tools/query_breakdown/src/main/java/com/google/bigquery/ReplacedComponent.java
```diff
@@ -0,0 +1,46 @@
+package com.google.bigquery;
+
+/**
+ * This class captures the replaced query, original component, and what it is replaced with
+ */
+public class ReplacedComponent {
+  private String query;
```
AayushBhatnagar, line 7: final modifier??

tools/query_breakdown/src/main/java/com/google/bigquery/Node.java
subinbean: Got it, changed it for the node class as well as the querybreakdown class!

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
subinbean: Ah! I mentioned above in the PR request that Exceptions will be dealt with in future PR's: for the next few PR's I'm focusing on the main features. Will refactor the exceptions in the future, leaving this comment open for now for future reference.

tools/query_breakdown/src/main/java/com/google/bigquery/ReplacedComponent.java
subinbean: Made all the fields final since replacedcomponent object will never change it's parameters

PR #122, opened 2020-07-28
SQL Extraction: vscode show progress

creativitRy:
A few days ago (#102), I added `--progress` to the CLI, which prints progress messages to STDERR. Now, I intercept those messages asynchronously in the VS Code extension, and I update the progress bar more naturally.

One complication was that STDERR data did not arrive in chunks of newlines but instead arrived in random lengths. To handle this, I batch all data into one string until a newline is found. Apparently, string concatenation is fast, and there's no StringBuilder class in the language libraries.


tools/vscode_sql_extraction/src/sql_extraction_runner.ts
```diff
@@ -54,7 +55,26 @@ export class SqlExtractionRunner {
       process.stdout!.on('data', data => {
         json += data;
       });
-      // todo: track progress
+      process.stderr!.on('data', data => {
+        errMsg += data;
+        let index = errMsg.indexOf('\n');
+        while (index >= 0) {
```
feiling, line 61: Seems easier to handle multiple '\n' by something like this:

```
if (index >= 0) {
  var lines = errMsg.split('\n');

  // assign last line to errMsg
  errMsg = lines[lines.length - 1];

  // display other lines
  for (var i = 0; i < line.length - 1; ++i)  {
    progress.report({
              message: line[i],
              increment: isNaN(percent) ? undefined : percent,
     });
  }
}
```

tools/vscode_sql_extraction/src/sql_extraction_runner.ts
```diff
@@ -41,6 +41,7 @@ export class SqlExtractionRunner {
 
     return new Promise<Query[]>((resolve, reject) => {
       let json = '';
+      let errMsg = '';
       const process = execFile(this.execPath, args)
```
feiling, line 45: There might be a way to disable buffering of stderr when you start a process. If so, implement that in another pull request.

tools/vscode_sql_extraction/src/sql_extraction_runner.ts
creativitRy: Great suggestion, thanks!

tools/vscode_sql_extraction/src/sql_extraction_runner.ts
creativitRy: Hmm yeah I'll look into it.

PR #123, opened 2020-07-29
Query_Breakdown: Program now executable and works correctly with deletion & correct error tracking 

subinbean:
### **TL;DR**
- We now have a working, executable program! The program works end-to-end with maven/cli and returns the correct error/error-location for deletion only cases. A few simple tests are included in InputReaderTest, LocationTrackerTest, and QueryBreakdownTest to ensure correctness. A notable case is: 

`BLAH SELECT a FROM A;`
 `BLAH SELECT b FROM B GROUP BLAH BY b`

now highlights errors as we want it to. 
- Files to look closely at: InputReader, LocationTracker, QueryBreakdown, Main

### **Specifics**
- InputReader as well as how we store the queries internally has completely changed: before, we maintained internally an invariant that every query would be single line and each new query would be separated by a new line. Now, the program just processes the input as is: this has allowed for multi-line queries to be processed seamlessly--simple tests have been added to ensure this. What is also nice is that this redesign got rid of the "semicolons in strings" problem. 

**LocationTracker is now implemented and can successfully track the original location of the error!**

1. A 2-D Arraylist of integers is used to keep track of the original positions. The indexing system is different (since line/columns start at 1) and therefore must be switched around accordingly. 
2. Deletion works by eliminating the deleted characters from the LocationTracker's position instance: essentially, position always mimics the current string but with the original locations: 

Input: `BLAH SELECT b FROM B GROUP BLAH BY b` (position: [[1, 2, 3, 4, ... ]]) 
Output: `SELECT b FROM B GROUP BY b`

When we delete BLAH, the intermediate query will be 
`SELECT b FROM B GROUP BLAH BY b` (position: [[5, 6, 7, 8, ...]])

Therefore, when we spot the error in BLAH after GROUP, the start/end column number will be wrong if we get it from the parser's error message. By deleting from the LocationTracker's position 2-D array the first 4 entries when we delete the first BLAH, however, we can use the error message's start/end column to get the original location since it's correctly shifted left by 4. 

3. We need a deep copy to be passed around the recursion since every path must have a unique copy. Otherwise, the location will not be correctly maintained
4. We initialize a LocationTracker when we first parse the input file through InputReader. Then, an instance is relayed from the main function to the run function, then to the loop. It is best for the LocationTracker to be initialized when reading the input first as the original locations are present there. The correct invariants for LocationTracker are maintained by the loop. 
- Minor change in deletion: we no longer take away extra spaces

### **Next Steps**
- Next PR: error-tracking working with Replacement
- Next, Next PR: VS Code Extension Draft 
- Ahead: VS Code Extension integration (including output format control through flags), Replacement logic improvement, Stabilization/Bug Fixes (private/public modifications, exception redesign, runtime limitation, etc.)


tools/query_breakdown/src/main/java/com/google/bigquery/LocationTracker.java
```diff
@@ -0,0 +1,92 @@
...
+     lt.remove(line, startColumn);
+    }
+    return lt;
+  }
+
+  /**
+   * To be implemented
+   */
+  public LocationTracker replace() {
+    return null;
```
feiling, line 72: throw UnsupportedOperationExceptions instead of return null.

tools/query_breakdown/src/main/java/com/google/bigquery/InputReader.java
```diff
@@ -12,31 +11,62 @@
  */
 public class InputReader {
 
+  private LocationTracker lt;
```
feiling, line 14: Just use name `locationTracker`.

tools/query_breakdown/src/main/java/com/google/bigquery/LocationTracker.java
```diff
@@ -0,0 +1,92 @@
+package com.google.bigquery;
+
+import java.util.ArrayList;
+
+/**
+ * This class tracks the original location of components in the query, thereby making sure that
+ * the error locations are correctly represented.
+ */
+public class LocationTracker {
```
feiling, line 9: It's not clear to me how LocationTracker works. It seems that for each pair of (line, column), `location.get(line - 1).get(column-1)` will return the location. But I don't understand what the location value represents. It's an integer, so I guess it mean the index in the input string?

subinbean: Ah! So for each pair of (line, column), it is initialized in position at the (line - 1)th arraylist and (column - 1)th element of the arraylist as the integer column - 1. Since the line number won't ever change (but the column number will change constantly), we simply keep track of the original column number as the integer in the double arraylist

tools/query_breakdown/src/main/java/com/google/bigquery/InputReader.java
subinbean: should I change all the lt's to locationTracker?

feiling: yes

subinbean: fixed!

tools/query_breakdown/src/main/java/com/google/bigquery/LocationTracker.java
feiling: Some question:
- Where is `location.get(x).get(y)` get assigned? 
- `ArrayList.add(x)` appends the value `x` to the array. It doesn't assign a value to `array[x]`. So in the method of  `public void add(int x, int y)`,  `location.get(x - 1).add(y)` doesn't seem right.

tools/query_breakdown/src/test/java/com/google/bigquery/LocationTrackerTest.java
```diff
@@ -0,0 +1,38 @@
...
+public class LocationTrackerTest {
+
+  @Test
+  public void locationTrackerTestInitializedCorrectly() throws IOException {
+    InputReader ir = new InputReader();
+    String absPath = new File("").getAbsolutePath();
+    ir.readInput(absPath + "/src/test/java/com/google/bigquery"
+            + "/InputTestFiles/multipleLineQuery.txt");
+    assertEquals(4, ir.getLocationTracker().getOriginalPosition(3, 4));
+  }
```
feiling, line 18: Unit tests should be self-contained, i.e. they don't rely on other data. So we shouldn't rely on the file `multipleLineQuery.txt`. Please add a method to `InputReader` such as `readFromString(string input)`.

tools/query_breakdown/src/main/java/com/google/bigquery/LocationTracker.java
subinbean: yup! more documentation is now added

tools/query_breakdown/src/test/java/com/google/bigquery/LocationTrackerTest.java
subinbean: Just added the method in inputreader and changed the tests accordingly

PR #126, opened 2020-07-31
VSCode_Query_Breakdown: Project Initialized 

subinbean:
Project structure set up for vs code extension--simple hello world runs 

P.S: Ignore all the QueryBreakdown related commits/files. This happened because I branched off from the milestone1_4 branch (the one where the PR is still open) so the old commits are present.


subinbean:
2 PR's were open at once and caused unnecessary complications

PR #129, opened 2020-07-31
SQL Extraction: python analyzer

creativitRy:
Support embedded SQL query extraction from python files. The python Antlr grammar file uses the MIT license. The grammar is for python3, but I don't think there are any backwards-incompatible grammar changes from python2 to python3.


tools/sql_extraction/src/test/kotlin/PythonAnalysisTest.kt
```diff
@@ -0,0 +1,66 @@
...
+    @Test
+    fun `literal concatenation`() {
+        val program = """
+import something
+something.some_method("hello" + "world")
+"""
+
+        val result = analyze(program)
+
+        assertCorrect(result, "(helloworld)")
```
feiling, line 42: How come the result is "(helloworld)" and not "helloworld"?

tools/sql_extraction/src/test/kotlin/PythonAnalysisTest.kt
```diff
@@ -0,0 +1,66 @@
...
+import something
+a = "test"
+something.some_method(a)
+"""
+
+        val result = analyze(program)
+
+        assertCorrect(result, "test")
+    }
+
```
feiling, line 57: Does the analyzer support multi-line strings?  If so, please add a unit test. If not, pls add support in another PR.

tools/sql_extraction/src/test/kotlin/PythonAnalysisTest.kt
creativitRy: For any complex query fragments, I added parentheses around the internal fragments (https://github.com/GoogleCloudPlatform/bigquery-utils/blob/master/tools/sql_extraction/src/main/kotlin/output/QueryFragment.kt#L74-L78).

Would it be a good idea to trim them off at the end of the recursion?

feiling: I see. That's fine. You can keep it.

tools/sql_extraction/src/test/kotlin/PythonAnalysisTest.kt
creativitRy: Good idea! I'll add tests for them

PR #131, opened 2020-08-03
Template Based Query Generation: Query generation (Merge Fix)

spoiledhua:
-This PR just fixes a merge error that occurred in the previous PR.


tools/template_based_query_generation/src/main/java/QueryGenerator.java
```diff
@@ -1,31 +1,146 @@
...
+import java.util.*;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Class that parses config file and creates queries from markov chain
  */
 public class QueryGenerator {
 
+	private final String filePathConfigDDL = "./src/main/resources/user_config/ddl.json";
```
Bei95, line 23: Please fix the indentation of this file to use 2 space of tab

tools/template_based_query_generation/src/main/java/QueryGenerator.java
```diff
@@ -1,31 +1,146 @@
...
+	private final String filePathConfigDML = "./src/main/resources/user_config/dml.json";
+	private final String filePathConfigDQL = "./src/main/resources/user_config/dql.json";
+	private final String filePathDependenciesRoot = "./src/main/resources/dialect_config/root_dependencies.json";
+	private final String filePathDependenciesDDL = "./src/main/resources/dialect_config/ddl_dependencies.json";
```
Bei95, line 27: As discussed offline, please consider generalize this file to easier add configs for other dialects.

tools/template_based_query_generation/src/main/java/parser/Pair.java
```diff
@@ -0,0 +1,16 @@
+package parser;
+
+public class Pair<F, S> {
```
Bei95, line 3: Java has it's own pair class. Any reason to create another pair class here?
https://docs.oracle.com/javase/8/javafx/api/javafx/util/Pair.html

tools/template_based_query_generation/src/main/java/QueryGenerator.java
AllenWang314: Resolved!

tools/template_based_query_generation/src/main/java/parser/Pair.java
AllenWang314: For some reason, Intellij + maven is giving me difficulties with javafx (this was why I decided to do it myself originally). I have deleted the class and used import internal.net.http.common.Pair instead

AllenWang314: So turns out that package isn't that great to work with either. This one was the one that worked in the end: org.apache.commons.lang3.tuple.MutablePair

PR #136, opened 2020-08-05
SQL Extraction: analyze java annotations

creativitRy:
A lot of Java frameworks use method annotations as a way of running SQL queries, and this PR allows analysis within annotation arguments.

Example from the framework JDBI (https://jdbi.org/#__sqlupdate):
```java
public interface UserDao {
  @SqlUpdate("insert into users (id, name) values (?, ?)")
  void insert(long id, String name);
}
```

In this case, the annotation would be considered a usage.


tools/sql_extraction/src/main/kotlin/DataFlowEngine.kt
```diff
@@ -30,6 +30,25 @@ class DataFlowEngine(private val environment: Environment = Environment()) {
...
+     *
+     * @param[visitArguments] Potential visit for all annotation arguments.
+     */
+    fun visitAnnotation(location: Location, visitArguments: () -> Unit) {
+        val prevQuery = currentQuery
+
+        currentQuery = null
+        visitArguments()
+
+        if (currentQuery != null) {
```
feiling, line 45: For the sample test 
```
   @SqlUpdate("insert into users (id, name) values (?, ?)")
    void insert(long id, String name);
```

What is the value of `currentQuery` here?

creativitRy: It would be the QueryFragment corresponding to the following json:
```javascript
{
  "count": "SINGLE",
  "location": {
    "startLine": 1,
    "startColumn": 17, // index of 'i' right after the first quote
    "endLine": 1,
    "endColumn": 58 // index of ')' right before the second quote
  },
  "literal": "insert into users (id, name) values (?, ?)" // quotes are not included
}
```

creativitRy: `CurrentQuery` will only be null if no string literals or variables holding query fragments are found in the annotation arguments.

PR #138, opened 2020-08-05
SQL Classifier: Added SQL classification

noah-kuo:
- This PR adds binary JAR files, which are the dialect parsers for Calcite. These should be REMOVED once these are published.


tools/unsupervised_dataset/sql_classifier/parserc/src/main/java/Parserc.java
```diff
@@ -0,0 +1,156 @@
...
+import org.apache.calcite.sql.parser.postgresql.PostgreSQLParserImpl;
+
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.text.SimpleDateFormat;
+import java.util.List;
+import java.util.Date;
+
+public class Parserc {
```
yzhvictor, line 17: Why it's called like this? Should be Classifier or something similar?

tools/unsupervised_dataset/sql_classifier/parserc/src/main/java/Parserc.java
```diff
@@ -0,0 +1,156 @@
...
+        if (args.length == 0) {
+            System.out.println("Please provide a CSV file.");
+            return;
+        }
+        List<String[]> allData = readCSV(args[0]);
+        if (allData == null) {
+            return;
+        }
+        SqlParser.Config[] parserConfigs = new SqlParser.Config[4];
+        parserConfigs[0] = SqlParser.configBuilder().setParserFactory(DefaultDialectParserImpl.FACTORY).build();
```
yzhvictor, line 34: Instead of adding it one by one, is it possible to do it in a loop to add all dialects Calcite has. Therefore, when Calcite has new dialect, it will be automatically included?

tools/unsupervised_dataset/sql_classifier/parserc/src/main/java/Parserc.java
```diff
@@ -0,0 +1,156 @@
...
+     * @return Contents of the CSV file
+     */
+    static List<String[]> readCSV(String filename) {
+        try {
+            FileReader filereader = new FileReader(filename);
+            CSVReader csvReader = new CSVReaderBuilder(filereader).withSkipLines(1).build();
+            return csvReader.readAll();
+        }
+        catch (Exception e) {
+            System.out.println(e);
```
yzhvictor, line 78: consider using logger instead of system.out.print

tools/unsupervised_dataset/sql_classifier/parserc/src/main/java/Parserc.java
```diff
@@ -0,0 +1,156 @@
...
+            return null;
```
yzhvictor, line 79: null may cause issues if other parts didn't handle correctly, consider using Optional<List<String[]>> as return type.

tools/unsupervised_dataset/sql_classifier/parserc/src/main/java/Parserc.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /*
+     * Classifies a single query using different SQL parsers.
+     *
+     * @param query The query to be classified
+     * @param parserConfigs The parsers for each of the different dialects
+     * @return A boolean array, with true values if the query can be classified in a dialect and false otherwise
+     */
+    static boolean[] classifyQuery(String query, SqlParser.Config[] parserConfigs) {
```
yzhvictor, line 90: will a map be more explicit? key can be dialect, value is true or false?

tools/unsupervised_dataset/sql_classifier/parserc/src/main/java/Parserc.java
noah-kuo: I'm not sure if this is possible, especially with the workaround we are using since we cannot get some of the dependencies online yet? I think this would definitely make the most sense when we have the capabilities, so this can be a change once it is available.

PR #139, opened 2020-08-06
Query Verification: Comparing query results

krishsuchdev:
Added the logic for converting BQ data types and for comparing the results of queries.


tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -112,5 +117,43 @@ public void verifyDataAware() throws InterruptedException {
...
+                    }
+                }
+
+                // Any leftover rows in the original results without a match are classified as missing in migrated results
+                missingResults.addAll(missingResultsSet);
+            }
+
+            return ResultDifferences.create(extraResults, missingResults);
+        } else {
+            return null;
```
yzhvictor, line 155: return null may cause unexpected issue if other parts didn't handle carefully. Consider using Optional<ResultDifferences> as return type

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
```diff
@@ -279,8 +281,55 @@ public String getName() {
             StandardSQLTypeName type = fields.get(i).getType().getStandardType();
 
             Object result;
-            // TODO Convert value to appropriate Java type
-            result = value.getStringValue();
+            try {
+                switch (type) {
```
yzhvictor, line 285: Do we need to have a default to at least put some warnings for unsupported types.

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
```diff
@@ -279,8 +281,55 @@ public String getName() {
...
+                    case TIMESTAMP:
+                        // returns Date
+                        result = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss.SSSSSS zzz").parse(value.getStringValue());
+                        break;
+                    case STRING: default:
+                        // returns String
+                        result = value.getStringValue();
+                }
+            } catch (ParseException e) {
+                result = null;
```
yzhvictor, line 331: use Optional.empty()

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -72,8 +74,11 @@ public void verifyDataFree() throws InterruptedException {
     /**
      * Verifies migrated query by sending query jobs to BQ and TD to check for differences in the query results.
      */
-    public void verifyDataAware() throws InterruptedException {
+    public void verifyDataAware() throws Exception {
         List<QueryJobResults> migratedResults = migratedInstance.runQueries();
+        List<QueryJobResults> originalResults = originalInstance.runQueries();
```
yzhvictor, line 79: will run queries keep the same order in 2 databases?

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
krishsuchdev: I have a default case that converts the value to a string to use for the other types not listed. Should I still add a warning?

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
krishsuchdev: Assuming that the 2 query files have the queries in the same order, `migratedResults` and `originalResults` should retain that order for the query results

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
```diff
@@ -279,8 +281,55 @@ public String getName() {
...
+                        result = value.getLongValue();
+                        break;
+                    case NUMERIC:
+                        // returns BigDecimal
+                        result = value.getNumericValue();
+                        break;
+                    case STRUCT:
+                        FieldList subFields = fields.get(i).getSubFields();
+                        FieldValueList subValues = value.getRecordValue();
+                        List<Object> subResults = parseResults(subValues, subFields);
```
Luminarys, line 305: I think you can combine this line like `result = parseResult(subValues, subFields)`.

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
```diff
@@ -279,8 +281,55 @@ public String getName() {
...
+                        break;
+                    case TIME:
+                        // returns Date
+                        result = new SimpleDateFormat("hh:mm:ss.SSSSSS").parse(value.getStringValue());
+                        break;
+                    case TIMESTAMP:
+                        // returns Date
+                        result = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss.SSSSSS zzz").parse(value.getStringValue());
+                        break;
+                    case STRING: default:
```
Luminarys, line 326: I would recommend making these separate cases and leaving a comment that we choose to use the String representation if the type is not known otherwise.

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
```diff
@@ -279,8 +281,55 @@ public String getName() {
             StandardSQLTypeName type = fields.get(i).getType().getStandardType();
 
             Object result;
-            // TODO Convert value to appropriate Java type
-            result = value.getStringValue();
+            try {
+                switch (type) {
+                    case BOOL:
+                        // returns boolean
```
Luminarys, line 287: I think we can remove all these `// returns` comments, the type information is already present

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
```diff
@@ -112,5 +117,43 @@ public void verifyDataAware() throws InterruptedException {
...
 
+    /**
+     * Finds extra and missing results by locating the differences between the results.
+     * @param migratedResults Parsed results returned from BQ
+     * @param originalResults Parsed results returned from original data warehouse service
+     * @return Differences classified as either extra or missing from migrated results.
+     */
+    public ResultDifferences compareResults(List<QueryJobResults> migratedResults, List<QueryJobResults> originalResults) {
+        // Check if same amount of queries were run
+        if (migratedResults.size() == originalResults.size()) {
```
Luminarys, line 128: Why does this have to be an if condition? I would assume we'd want to return information around differences in row counts etc.

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
Luminarys: Seconding this. I also wonder why we even need this else branch since I think we should be able to produce a result difference no matter what the size of each result set is.

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
krishsuchdev: This is actually just checking if the same number of query statements were run. If they are not equal, that means either `migratedResults` or `originalResults` has the results of queries that aren't in the other and it'd be difficult to determine which ones those are

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
krishsuchdev: Updated. The condition was just checking if the same number of query statements were run. If they weren't, it meant one of the parameters had results from a query that wasn't ran in the other

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
```diff
@@ -279,8 +281,46 @@ public String getName() {
...
+                        result = new SimpleDateFormat("yyyy-MM-dd'T'hh:mm:ss.SSSSSS").parse(value.getStringValue());
+                        break;
+                    case TIME:
+                        result = new SimpleDateFormat("hh:mm:ss.SSSSSS").parse(value.getStringValue());
+                        break;
+                    case TIMESTAMP:
+                        result = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss.SSSSSS zzz").parse(value.getStringValue());
+                        break;
+                    case STRING:
+                        result = value.getStringValue();
```
Luminarys, line 316: Don't forget the break

tools/query_verification/src/main/java/com/google/bigquery/QueryVerifier.java
Luminarys: I see, makes sense then. Thanks!

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
```diff
@@ -279,8 +281,46 @@ public String getName() {
...
+                    default:
+                        // Handle unknown/unsupported types as String
+                        result = value.getStringValue();
+                }
+            } catch (ParseException e) {
+                result = Optional.empty();
```
Luminarys, line 322: Zihao did you want each object to be Optional or did you want an empty Optional added like this? I actually think null is ok in this situation.

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
krishsuchdev: Ah, good catch!

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
yzhvictor: I think Optional.empty() looks good. My impression is java guide suggests to use optional whenever applicable.

tools/query_verification/src/main/java/com/google/bigquery/BigQueryManager.java
yzhvictor: Adding a warning sounds good, since it can help us identify the unsupported types.

PR #142, opened 2020-08-07
Query_Breakdown: Tool now works with Replacement

subinbean:
- Integrated LocationTracker with Replacement
- The tool's main algorithm components (run and loop method) now work with replacement
- Implemented a stack to change the output ordering in a more intuitive way (top down rather than bottom up)
- Revised output-printing code to better support replacement


tools/query_breakdown/src/main/java/com/google/bigquery/LocationTracker.java
```diff
@@ -40,6 +40,14 @@ public void addLine() {
...
   }
 
+  /**
+   * This method inserts an element in the middle of the location ArrayList to replicate
+   * replacement
+   */
+  public void add(int x, int y, int numberToAdd) {
+    location.get(x - 1).add(y, numberToAdd);
+  }
+
```
feiling, line 50: Seem this method, and `change()`, are only used by `replace()`. If that's the case, you should make them private.

tools/query_breakdown/src/main/java/com/google/bigquery/LocationTracker.java
```diff
@@ -51,6 +59,17 @@ public void remove(int x, int y) {
     }
   }
 
+  /**
+   * This method changes an entry from location specified by (x, y) in the original query.
+   */
+  public void change(int x, int y, int changeTo) {
```
feiling, line 65: Where is this method used?

tools/query_breakdown/src/main/java/com/google/bigquery/LocationTracker.java
subinbean: Got it, yeah it's definitely more of a helper function--changed!

tools/query_breakdown/src/main/java/com/google/bigquery/LocationTracker.java
subinbean: function now deleted!

PR #146, opened 2020-08-11
SQL Extraction: parallelize analysis using coroutines

creativitRy:
I found out that Kotlin has coroutines just like go's goroutines! They are lightweight and multiplexes to the available number of hardware threads.

Since I already designed the program to have no global mutations (each file is analyzed in isolation), I managed to multithread the entire operation. The majority of the code changes comes from moving parts of earlier code into a method.

On a test repository:
- average timing without --parallel: 29.41 seconds
- average timing with --parallel (2 cores 4 logical processors): 20.79 seconds
- speedup: x1.4145


creativitRy:
I tried running the test repo with an 8 core 2 threads machine:
- average timing: 11.14 seconds
- speedup over sequential: x2.6392

PR #147, opened 2020-08-12
Query_Breakdown/VSCode_Query_Breakdown: Integration Complete

subinbean:
### Integration between the VS Code Frontend and CLI Backend is complete 

- JSON flag -j is added, so the output is now available in a JSON format
- The program can now generate a stand-alone jar file which the VS Code Frontend runs 
- query_breakdown_runner.ts is created to write the execute function that returns a promise (for async) after executing the backend and parsing the results into JSON objects
- resultJson.ts is created to make the results from the backend into a structured JSON format 
- extension.ts is updated to support the integration. It asynchronously runs query_breakdown_runner and receives the results in a JSON array format
- more relevant documentation can be found in the code 

P.S: the code complies to Google's TS style according to gts (open source tool for Google's typescript style)


tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
```diff
@@ -2,34 +2,31 @@
...
 /**
  * This file is the main file for the command line tool.
- * Usage: query_breakdown -i <PATH> [-o <PATH>] [-l <INTEGER>]
+ * Usage: query_breakdown -i <PATH> [-j] [-l <INTEGER>]
  * -i, --inputFile, PATH: this command specifies the path to the file containing queries to be
  *                    inputted into the tool. It is therefore mandatory
- * -o, --outputFile, PATH: this command specifies the path to the file that the tool can write
- *                    its results to. If not specified, the tool will simply print results on the
- *                    console. It is therefore optional
+ * -j, --json, PATH: this command specifies whether the program should output the results in a
```
feiling, line 14: Seems `, PATH` should not be here?

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
```diff
@@ -2,34 +2,31 @@
...
+ *                   json format. It is therefore optional
  * -l, --limit, PATH: this command specifies the path to an integer that the tool takes as a
```
feiling, line 16: Seems `PATH` should be `INTEGER`

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
```diff
@@ -62,23 +64,51 @@ public void run(String originalQuery, String outputFile, int errorLimit,
...
-                + "Start Column %3$s, End Column %4$s, %5$s", current.getStartLine(),
-            current.getEndLine(), current.getStartColumn(), current.getEndColumn(),
-            current.getErrorHandlingType()));
+    
+    if (jsonOutput) {
+      JSONArray jsonArray = new JSONArray();
+      while(!stack.empty()) {
+        current = stack.pop();
+        JSONObject errorPosition = new JSONObject();
+        errorPosition.put("startLine", current.getStartLine());
```
feiling, line 73: You should be able to simplify this by using a library like Jackson (https://en.wikipedia.org/wiki/Jackson_(API)). You can choose to do that in a future PR.

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
subinbean: yup, changed

tools/query_breakdown/src/main/java/com/google/bigquery/Main.java
subinbean: yup, changed!

tools/query_breakdown/src/main/java/com/google/bigquery/QueryBreakdown.java
subinbean: Got it, yup. Right now I might still add more to the JSON schema so I will address this in a future PR!

PR #155, opened 2020-08-19
Auto Query Fixer: use Regex to locate the incorrect table when fixing the TableNotFound error

mingen-pan:
Modify a member function in `TableNotFoundFixer` to use regex to locate the incorrect table.

The new method can handle the table names like `a.b.c`, `a`.b.c, `a.b`.c, `a`.`b`.`c`, a.b.c, and etc.


tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/entity/TokenImpl.java
```diff
@@ -13,8 +13,11 @@ public TokenImpl (Token token) {
     this.token = token;
   }
 
-  @Override public int getKind() {
-    return token.kind;
+  @Override public Kind getKind() {
+    if (token.kind == 788) {
```
kikkyo, line 17: 788? i think it should corresponding to the index defined in the kind enum?

tools/automatic_query_fixer/src/test/java/com/google/cloud/bigquery/utils/queryfixer/FixerTest.java
```diff
@@ -47,6 +47,7 @@ public void setup() {
     fixerFactory = new FixerFactory(tokenProcessor, bigQueryServiceMock);
   }
 
+  //TODO: More test cases are need to test different table name, like `a.b`.c, `a.b.c`, `a`.`b`.c, and a.b.c.
```
kikkyo, line 50: i think the most common case would be 'datasetid.tableid', please make sure we handle this case.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/fixer/TableNotFoundFixer.java
```diff
@@ -100,24 +106,32 @@ private String constructFullTableName(String projectId, String datasetId, String
...
-
-    // Since the TableNotFound error has no position info, this method will convert the index to the
-    // position and assign to the `err`.
-    Position position = queryPositionConverter.indexToPos(index);
-    this.err.setErrorPosition(position);
-    return index;
+  private SubstringView findSubstringViewOfIncorrectTable(TableId fullTableId) {
+    String regex =
+        String.format(
+            "`?%s`?.`?%s`?.`?%s`?",
```
kikkyo, line 112: "?" would match one or more times, what is it is "datasetid.tableid" without "`"?
how about "`datasetid.tableid`" which is the most common case.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/entity/TokenImpl.java
mingen-pan: Calcite defines hundreds of token kinds, so number 788 indeed means an ordinary identifier. Anyway this token class will be deprecated very soon once the ZetaSQL server is setup and a ZetaSQL token is introduced.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/entity/TokenImpl.java
```diff
@@ -13,8 +13,11 @@ public TokenImpl (Token token) {
     this.token = token;
   }
 
-  @Override public int getKind() {
-    return token.kind;
+  @Override public Kind getKind() {
+    if (token.kind == 786 || token.kind == 788) {
```
kikkyo, line 17: I forgot maybe you explained last time why it is 786?

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/fixer/TableNotFoundFixer.java
```diff
@@ -100,24 +107,43 @@ private String constructFullTableName(String projectId, String datasetId, String
...
-
-    // Since the TableNotFound error has no position info, this method will convert the index to the
-    // position and assign to the `err`.
-    Position position = queryPositionConverter.indexToPos(index);
-    this.err.setErrorPosition(position);
-    return index;
+  private SubstringView findSubstringViewOfIncorrectTable(TableId fullTableId) {
+    String regex;
+    // If the project ID is the default one, then the actual table may not include project ID,
+    // which looks like "dataset.table".
```
kikkyo, line 113: Theoretically I think we should handle single 'table' too? We can add TODO here.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/entity/TokenImpl.java
mingen-pan: 786 is the identifier with backticks, but 788 is an identifier without any enclosing.

mingen-pan: They will be deprecated very soon. Probably next a few PRs when the ZetaSQL is used.

tools/automatic_query_fixer/src/main/java/com/google/cloud/bigquery/utils/queryfixer/fixer/TableNotFoundFixer.java
mingen-pan: Yes, if a user specifies the default dataset in the input options. TODO is added.

PR #161, opened 2020-08-22
Query_Breakdown: New Query Breaking Down Logic Implemented, Multi-line deletion implemented, Replacement Logic randomized

subinbean:
### New Query-Breaking Down Logic
- To optimize for runtime, I decided to modify the design and create a separate tree for each query rather than for the entire document. I do this by creating a new QueryBreakdown instance as well as a new LocationTracker instance for each query. The queries are broken down during input parsing. 
- Location tracking was very difficult, but it was fixed by implementing a pair design where it is a double arraylist of pairs that denote the original position. By doing so, we were able to maintain the original mappings despite breaking down the queries. (We could do this internally by keeping a localLine field)
- The way we output has also changed: now, we return a list of nodes and aggregate them accordingly in the main function to output the final result.
- All the tests have been modified to serve this new design 

### Multi-line Deletion
- A major design flaw has been discovered, which is that errors do not always occur on the same line. This assumption changed up the deletion mechanism design as well as the location-tracking design. 
- There were numerous edge cases that had implications for location-tracking, such as the case where there were middle lines, the endline was all deleted, the first line was all deleted, etc. The hardest thing was that new line characters would be deleted as well, messing up the location tracking mechanism. This was dealt with and tested. 

### Replacement logic randomized, misc. bugs fixed 
- The replacement logic was re-written to be randomized instead of choosing the first three.
- Other issues were dealt with for stabilization, such as having no replacement options, there being a <EOF> token at the end of the file, etc.


tools/query_breakdown/src/main/java/com/google/bigquery/ReplacementLogic.java
```diff
@@ -9,23 +10,50 @@
...
+   * replaced with. Returns a list of components that it recommends. The default method of
+   * choosing recommendations is through randomization.
    *
    * We choose ArrayLists as the data structure because we want to impose a certain ordering
    * with the recommendations: certain recommendations should be "better" than others
+   *
+   * n is an integer that controls the number of replacement options
    */
   public static ArrayList<String> replace (String component, ArrayList<String> options) {
-    // simply returns the first n options
```
feiling, line 18: Please keep the comment so it's clear what's the purpose of `n`.

tools/query_breakdown/src/main/java/com/google/bigquery/ReplacementLogic.java
```diff
@@ -9,23 +10,50 @@
...
+      else {
+        result.add(options.get(random));
+        seen.add(random);
+      }
     }
     return result;
   }
+
+  // helper function that returns a random integer between 0 - n
+  private static int randomNumber(int n) {
```
feiling, line 56: You should use `Random.nextInt()` (https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#nextInt--)

tools/query_breakdown/src/main/java/com/google/bigquery/ReplacementLogic.java
subinbean: addressed in the next PR!

PR #162, opened 2020-08-26
Auto Query Fixer: implement Tokenize and some utils functions in ZetaSQL Helper server

mingen-pan:
Implement Tokenize and Token class
RPC call for Tokenize is set up.
Util function are developed for future features
Test case are included.


tools/zetasql_helper/zetasql_helper/local_service/BUILD
```diff
@@ -24,6 +25,19 @@ cc_grpc_library(
     deps = [":local_service_cc_proto"],
 )
 
+cc_library(
+    name = "local_service",
+    srcs = ["local_service.cc"],
+    hdrs = ["local_service.h"],
+    copts = ["-Wno-sign-compare"],
```
kikkyo, line 32: curious what is this used for?

tools/zetasql_helper/zetasql_helper/local_service/local_service.proto
```diff
@@ -11,6 +28,9 @@ service ZetaSqlHelperLocalService {
   rpc Hello(HelloRequest) returns (HelloResponse) {
   }
 
+  rpc Tokenize(TokenizeRequest) returns (TokenizeResponse) {
+  }
+
 }
 
 message HelloRequest {
```
kikkyo, line 36: we don't need this HelloRequest/Response right?

tools/zetasql_helper/zetasql_helper/local_service/BUILD
mingen-pan: I took this compiler options from the ZetaSQL BUILD file. I guess it asks the compiler not to care about the sign of data types.

tools/zetasql_helper/zetasql_helper/local_service/local_service.proto
mingen-pan: No we don't need it. I will remove it.

PR #163, opened 2020-08-27
Token generalization: Regex parsing and formatting

spoiledhua:
Finished the main code for interpreting custom regex, allowing a wider range of user input and query formatting. Queries are printed to standard output for now but will be written in data files.


tools/template_based_query_generation/src/test/java/graph/MarkovChainTest.java
```diff
@@ -64,7 +62,7 @@ public void test_randomWalk_anEdge() {
 	 * test for the randomWalk method when the graph consists of a bidirectional edge
 	 * an infinite loop happens in the graph.MarkovChain class, so we expect an OutOfMemoryError.class
 	 */
-	@Test
+	//@Test
```
Bei95, line 65: Is this intentional? If it is, please include some comments for it. Thx!

tools/template_based_query_generation/src/test/java/parser/KeywordsMappingTest.java
```diff
@@ -1,48 +1,10 @@
...
 import org.junit.jupiter.api.Test;
-import token.TokenInfo;
-
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertThrows;
```
Bei95, line 4: Please add comments & ToDo when the unit test is not provided.

tools/template_based_query_generation/src/main/java/token/QueryRegex.java
```diff
@@ -0,0 +1,115 @@
...
+
+import parser.Utils;
+import query.SkeletonPiece;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Random;
+
+public class QueryRegex {
```
Bei95, line 11: Looks like this class only contains one function, so probably it makes more sense to move this function to Query class?

tools/template_based_query_generation/src/test/java/graph/MarkovChainTest.java
spoiledhua: Yes, will add the unit tests today!

tools/template_based_query_generation/src/test/java/parser/KeywordsMappingTest.java
spoiledhua: Will add!

PR #178, opened 2020-10-05
Create get_table_ddl.sql

PraveenAkunuru:
Adding procedure to get ddl. 

Usage:
    call procedure.get_table_ddl('project.dataset.<pattern>%')
    call procedure.get_table_ddl('project:dataset.<pattern>%')
    call procedure.get_table_ddl('project.dataset.%') to get all ddl's in a dataset
 Project name is mandatory because we can always refer to tables in a Project different from where we are calling the function. At this time this procedure only generates ddl for BigQuery native tables. Does not provide ddl for External tables, views.

Including the prerequisite functions within this .sql file itself for ease of setup.

### DIFF
--- a/stored_procedures/get_table_ddl.sql
+++ b/stored_procedures/get_table_ddl.sql
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


danieldeleo:
Closing this PR since BigQuery now provides DDL in the [TABLES view](https://cloud.google.com/bigquery/docs/information-schema-tables#tables_view)

PR #182, opened 2020-10-23
gcs ingest cloud function

jaketf:
Adds cloud function to ingest batches of files from GCS with an opinionated storage layout, flexible configuration system and supporting load jobs and external queries as a method of ingest (see README for details).

Please let me know if adding BQ utility cloud functions like this would be an appropriate use of this repo and how best to hook my integration tests into CI on PRs.


cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,378 @@
...
+                f"query job {job.job_id} failed quickly: {job.errors}")
+        sleep(1)
+
+
+def flatten(arr: List[List[Any]]) -> List[Any]:
+    """Flatten list of lists to flat list of elements"""
+    return [j for i in arr for j in i]
+
+
+def load_batches(gcs, bq, gsurl, dest_table_ref):
```
jaketf, line 170: without changing the configuration interface we could optionally swap out this `load_batches` for creating a GCS to BQ Data Transfer Service config and submitting a new data transfer. This would have the advantage of relying on DTS's existing mechanics to orchestrate multiple loads to a staging table and atomically commit partitions to the destination table with a copy job. This has the advantage of atomicity of loading a partition but the disadvantage of a requiring a call to bigquerydatatransfer.googleapis.com which may prohibit use in environments that can't route this address due to lack of restricted VIP.

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,378 @@
...
+from urllib.parse import urlparse
+import logging
+from time import monotonic, sleep
+
+from google.api_core.client_info import ClientInfo
+from google.cloud import bigquery, storage
+from google.cloud.exceptions import NotFound
+
+# 15TB per BQ load job.
+MAX_BATCH_BYTES = 15 * 10 ** 12
```
jaketf, line 34: this should be configurable w/ environment variable for customers who have negotiated a higher per load job bytes limit.

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
ryanmcdowell: I think you should still be able to use BigQuery DTS GCS to BQ inside the VPC-SC. Not sure where the gaps are between transfer services but other customers have been running these types of jobs inside the VPC.

cloud_functions/gcs_event_based_ingest/README.md
```diff
@@ -0,0 +1,116 @@
...
+{
+    "sourceFormat": "AVRO",
+    "writeDisposition": "WRITE_TRUNCATE",
+    "schemaUpdateOptions": ["ALLOW_FILED_RELAXATION"]
+}
+```
+This configuration system gives us the ability to DRY up common defaults but
+override them at whatever level is appropriate as new cases come up.
+
+#### Transformation SQL
```
ryanmcdowell, line 80: Perhaps a notice that this will utilize the Query slots as opposed to the free load slot tier to execute would be good.

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/README.md
```diff
@@ -0,0 +1,73 @@
...
+From the `gcs_ocn_bq_ingest` dir simply run
+```bash
+pytest
+```
+
+## Limitations
+1. Cloud Functions have a 10 minute timeout. If the BQ load job takes too long
+ the data will still be ingested but the function may be marked in timeout state
+ rather than success and the function may not have the opportunity to produce
+ all the expected logs.
```
ryanmcdowell, line 65: I'm curious how you would suggest we track what data has been ingested given this limitation?

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,378 @@
+# Copyright 2018 Google LLC.
```
ryanmcdowell, line 1: 2020

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,378 @@
...
+# This may not be honored if longer than cloud function timeout
+# https://cloud.google.com/functions/docs/concepts/exec#timeout
+WAIT_FOR_JOB_SECONDS = 5
+
+
+def main(event: Dict, context):
+    # Set by Cloud Function Execution Environment
+    # https://cloud.google.com/functions/docs/env-var
+    project = getenv("GCP_PROJECT")
+
```
ryanmcdowell, line 72: Remove extra blank line

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,378 @@
...
+
+    bucket_id, object_id = parse_notification(event)
+
+    success_filename = getenv("SUCCESS_FILENAME", "_SUCCESS")
```
ryanmcdowell, line 76: Seems like this could be a variable defined above given `get_batches_for_prefix` depends on the same default

cloud_functions/gcs_event_based_ingest/tests/test_gcs_ocn_bq_ingest_it.py
```diff
@@ -0,0 +1,235 @@
+# Copyright 2019 Google LLC
```
ryanmcdowell, line 1: 2020

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,378 @@
...
+
+    # Exit eagerly if not a success file.
+    # TODO we can improve this with pub/sub message filtering once it supports
+    # a hasSuffix filter function (we can filter on hasSuffix successfile name)
+    #  https://cloud.google.com/pubsub/docs/filtering
+    if not object_id.endswith(f"/{success_filename}"):
```
ryanmcdowell, line 82: If a directory which has been previously loaded receives new data and a new `_SUCCESS` file, it seems like entire table will be reloaded. Is it safe to say this does not handle incremental loads? If so, perhaps the default disposition should be `WRITE_TRUNCTATE`

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/README.md
jaketf: I should update these docs.
Essentially the cloud function should exit quickly always (does not block on load job completion just wait n ~= 5 seconds to make sure it didn't fail quickly).
Tracking of "what data has been ingested" is all captured in INFORMATION_SCHEMA / Audit logs.
All jobs submitted will follow a naming convention and be labelled.

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
jaketf: I definitely want to handle incrementals.

I will add a note to the docs and a test for this as this will be a common use case. 
Incrementals should always happen in a prefix under the table prefix (e.g. a partition prefix) so that their behavior can be properly overwritten in a load.json .
e.g.
```
gs://bucket/dataset/table/_config/load.json          # load.json dictates WRITE_APPEND for historical
gs://bucket/dataset/table/parts/_config/load.json # load.json WRITE_TRUNCATE for all prefixes under parts/
```

However, if willing to pay for query slots for incrementals IMHO the most general way to solve incrementals should happen via external query as this provides richer DML semantics (e.g. `MERGE` into partition) than load job write disposition and access to the `_FILE_NAME` which can be used to. 

In the future as GCS > BQ DTS ongoing "mirroring" is supported, this asset would mostly be useful for orchestrating external query ingest.

jaketf:
TODO(@jaketf) make this more flexible for reading dataset_id, table_id, partition_id, batch_id from source uri / SUCCESS file path.

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
jaketf: @ryanmcdowell can you see if the docs and regex system I've added in latest commits addresses your incremental loads use case? (of course it will be better once I've added integration test for this pattern)

jaketf:
@danieldeleo @ryanmcdowell 
I believe I've addressed all feedback, though feel free to add more.
If you feel we'd like to accept this contribution I'd like to request access to the CI project OR could you set up two cloud build triggers to run `cloud_functions/gcs_event_based_ingest/cloudbuild.yaml` the CI I added in d2d4dc9: 
1. for PRs when files under `cloud_functions/gcs_event_based_ingest/*` are changed with exlcusion filter of `*.md`.
1. nightly (optional but would allow us to catch breaking python upgrades) 
I've done my best to add thorough documentation so anyone can understand how to run the tests locally or work with the cloudbuild and why certain decisions were made. I think this provides really nice example of good CI process for python cloud functions and I might port this to PSO github if we don't think it's appropriate here.
Please LMK if this is not how you would like to manage CI for this repo.

cloud_functions/gcs_event_based_ingest/Dockerfile.ci
```diff
@@ -0,0 +1,6 @@
+FROM python:3.8-slim
+COPY requirements.txt requirements-dev.txt .
+RUN pip3 install -r requirements-dev.txt
+WORKDIR /ci
+COPY . /ci/
+ENTRYPOINT ["pytest"]
```
ryanmcdowell, line 6: Trailing newline

cloud_functions/gcs_event_based_ingest/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Partitions
+
+#### Partition Table Decorators
+Note that if the directory immediately before the triggering successfile starts with
+a `$` it will be treated as a BigQuery Partition decorator for the destination table.
+
+This means for:
+```text
+gs://${BUCKET}/foo/bar/$20201026/_SUCCESS
```
ryanmcdowell, line 124: Why not just default this to the hive partitioning layout?

cloud_functions/gcs_event_based_ingest/README.md
```diff
@@ -0,0 +1,248 @@
...
+This can be helpful in reprocessing scenarios where you'd want to `WRITE_TRUNCATE`
+a partition that had some data quality issue.
+
+#### Hive Partitioning
+If your data will be uploaded to GCS from a hadoop system that uses the 
+[supported default hive partitioning](https://cloud.google.com/bigquery/docs/hive-partitioned-loads-gcs#supported_data_layouts)
+you can specify this in the [`hivePartitioningOptions`](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#hivepartitioningoptions)
+key of `load.json` for that table.
+
+Any non-trivial incremental loading to partitions should usually use the
```
ryanmcdowell, line 137: should but is not required correct?

cloud_functions/gcs_event_based_ingest/README.md
```diff
@@ -0,0 +1,248 @@
...
+   `region-us`.INFORMATION_SCHEMA.JOBS_BY_PROJECT
+WHERE
+   (SELECT value FROM UNNEST(labels) WHERE key = "component") = "event-based-gcs-ingest"
+```
+
+## Triggers
+
+### Pub/Sub Storage Notifications `_SUCCESS`
+1. Trigger on `_SUCCESS` File to load all other files in that directory.
+1. Trigger on non-`_SUCCESS` File will no-op
```
ryanmcdowell, line 200: I'm curious if there's an alternative way to implement this. For instance I see for incrementals we include batch id. But alternatively these semantics could be specified as:

Batch load if a success file is specified, otherwise load files as they land into GCS. If the user wants transactional semantics, they can use the success file method. The WRITE_DISPOSITION can be separately defined in the config such that either method could technically be incremental.

cloud_functions/gcs_event_based_ingest/tests/test_gcs_ocn_bq_ingest_it.py
```diff
@@ -0,0 +1,388 @@
...
+    def teardown():
+        bq.delete_table(table, not_found_ok=True)
+
+    request.addfinalizer(teardown)
+    return table
+
+
+@pytest.fixture(scope="function")
+@pytest.mark.usefixtures("gcs_bucket", "dest_dataset", "dest_table")
+def gcs_data(request, gcs_bucket, dest_dataset,
```
ryanmcdowell, line 111: Where is this referenced?

cloud_functions/gcs_event_based_ingest/README.md
jaketf: Interesting proposal.  
tl;dr this should be addressed w/ another tool and if we want to support it in this function it should be a separate PR IMHO.

I had a similar thought but decided not to implement it this way. Do you have a customer with actual requirements like this? I'm open to thinking about it more but specific requirements would be helpful.
e.g. 
- freshness requirements (aka why not use scheduled loads every 2-5 mins? could even use this solution and drop a success file and write truncate the partition every few mins / or use transnational external query for more specificity)
- query performance requirements (aka why not use external table?)
- duplicate loading of same file tolerance 
- how many files per day / hour
- "files not loaded sequentially" tolerance

not sure what kind of queries don't care about the above things except speculative results which could usually be served w/ external table.

IMHO this use cases of "load each file as it lands" are 90%  of the time either better served with external query the data directly OR pivoting publishers to streaming inserts.

I've originally avoided this because it will create an inefficient number of load jobs (resulting in inefficient number of storage sets IIUC which may damage performance of queries until some compaction of storage sets occurs) and I've been focused on scalability.
It will become much more likely to hit max load jobs per project / table https://cloud.google.com/bigquery/quotas#load_jobs.
This also makes the system harder to monitor / reason about.

Also this would complicate handling duplicate OCN pubsub messages as we'd have to use a more sophisticated de-duplication strategy than the one currently implemented (based on a single lock file). This will likely require a transactional database (cloud SQL, firestore, memorystore, bigtable). In my experience customers doing BQ migrations don't have a lot of GCP services approved by their infosec teams yet making additional complexity in the architecture a blocker to many orgs, not to mention adding cost / maintenance overhead of this metadata db to an otherwise cheap / managed solution.

cloud_functions/gcs_event_based_ingest/tests/test_gcs_ocn_bq_ingest_it.py
jaketf: It's a pytest fixture (sort of confusing dynamic syntax) used by these tests:

https://github.com/GoogleCloudPlatform/bigquery-utils/pull/182/files/4a345c8d3aa4068c1c9631cb052cd316f5039c85#diff-0e25c24817556a4a7ba570e46eebf9adf31ad1c091130b4045220a5c54c532f3R132

https://github.com/GoogleCloudPlatform/bigquery-utils/pull/182/files/4a345c8d3aa4068c1c9631cb052cd316f5039c85#diff-0e25c24817556a4a7ba570e46eebf9adf31ad1c091130b4045220a5c54c532f3R157

cloud_functions/gcs_event_based_ingest/README.md
jaketf: My thought is this default supports date, timestamp and integer range partition decorators and is explicit / matches BQ target syntax.
The hive partitioning is useful only to those following a supported folder structure.
In a future PR I can add more detailed docs and an integration test for how we'd suggest using this function with native BQ support for hive partitioning by specifying it in load.json at the table level.

The way the docs for BQ hive partitioning reads is that you can provide `dt=/val=` however most often in the wild (e.g. my current customer) I see data lakes of partitioned data in `table/yyyy/MM/dd/hh/` which is [unsupported by bq hive partitioning](https://cloud.google.com/bigquery/docs/hive-partitioned-loads-gcs#unsupported_data_layouts) because partition key name is not encoded in the path. Even for `yr=yyyy/m=MM/d=dd/hr=hh` (ala my first ad tech customer) this is "supported" but all of a sudden you can only partition on fields `yr`, `m`, `d`, or  `h` rather than the whole date time (which defeats the purpose). Often times there are other ETL process / consumers who need the data in this exact layout so the customer is not that flexible to change.

For these customers I'd recommend using external query loads and define a small udf to parse your org's partitioning / naming convention and constructing the timestamp / date time field you want to partition on. Alternatively they can fork this function for those use cases and add some extra parsing step on the partition extracted by regex to produce the partition decorator they desire for their destination table ref.

cloud_functions/gcs_event_based_ingest/README.md
jaketf: correct.

cloud_functions/gcs_event_based_ingest/Dockerfile.ci
jaketf: fixed also a good cue that this dockerfile should also be linted :) I've also added a hadolint step to cloudbuild.

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,459 @@
...
+from google.api_core.client_info import ClientInfo
+from google.api_core.exceptions import PreconditionFailed
+from google.cloud import bigquery, storage
+from google.cloud.exceptions import NotFound
+
+# 15TB per BQ load job.
+DEFAULT_MAX_BATCH_BYTES = str(15 * 10**12)
+MAX_URIS_PER_LOAD = 10**4
+
+DEFAULT_EXTERNAL_TABLE_DEFINITION = {"sourceFormat": "PARQUET"}
```
danieldeleo, line 37: Let's default this to CSV to align with the load job default

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,459 @@
...
+from urllib.parse import urlparse
+
+from google.api_core.client_info import ClientInfo
+from google.api_core.exceptions import PreconditionFailed
+from google.cloud import bigquery, storage
+from google.cloud.exceptions import NotFound
+
+# 15TB per BQ load job.
+DEFAULT_MAX_BATCH_BYTES = str(15 * 10**12)
+MAX_URIS_PER_LOAD = 10**4
```
danieldeleo, line 35: Nit, but perhaps rename to MAX_SOURCE_URIS_PER_LOAD and/or add to above comment the following link: https://cloud.google.com/bigquery/quotas#load_jobs

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,459 @@
...
+DEFAULT_EXTERNAL_TABLE_DEFINITION = {"sourceFormat": "PARQUET"}
+
+DEFAULT_JOB_LABELS = {
+    "component": "event-based-gcs-ingest",
+    "cloud-function-name": getenv("FUNCTION_NAME"),
+}
+
+BASE_LOAD_JOB_CONFIG = {
+    "sourceFormat": "CSV",
+    "fieldDelimiter": "|",
```
danieldeleo, line 46: bq tool uses comma as default field delimiter, let's align with that tool

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,459 @@
...
+    This hinges on a SQL query defined in GCS at _config/bq_transform.sql and
+    an external table definition _config/external.json (otherwise will assume
+    parquet external table)
+    """
+    external_table_config = read_gcs_file_if_exists(
+        gcs_client, f"{gsurl}_config/external.json")
+    if external_table_config:
+        external_table_def = json.loads(external_table_config)
+    else:
+        print(f"Falling back to default parquet external table."
```
danieldeleo, line 189: Update message to default CSV external table

cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/main.py
```diff
@@ -0,0 +1,459 @@
...
+
+    external_table_def["sourceUris"] = flatten(
+        get_batches_for_prefix(gcs_client, gsurl))
+    external_config = bigquery.ExternalConfig.from_api_repr(external_table_def)
+    job_config = bigquery.QueryJobConfig(
+        table_definitions={"temp_ext": external_config}, use_legacy_sql=False)
+    # for some reason query string literal wrapped in b''
+    rendered_query = str(
+        str(query).format(dest_dataset=dest_table_ref.dataset_id,
+                          dest_table=dest_table_ref.table_id))[2:-1]
```
danieldeleo, line 201: query is wrapped in b'' because it's a bytestring. Instead of [2:-1] just call --> query.decode('UTF-8')

jaketf: TIL thank you.

cloud_functions/gcs_event_based_ingest/Dockerfile.ci
```diff
@@ -1,6 +1,4 @@
 FROM python:3.8-slim
 COPY requirements.txt requirements-dev.txt ./
 RUN pip3 install -r requirements-dev.txt
-WORKDIR /ci
```
jaketf: cc: @danieldeleo FYI because we know cloud build we have the source cloned locally we can remove this step speeding up image build time significantly as the ci container doesn't need the source inside the container.

cloud_functions/gcs_event_based_ingest/tests/test_gcs_ocn_bq_ingest_it.py
```diff
@@ -374,15 +357,54 @@ def test_external_query_IT(bq, gcs_data, gcs_external_config, dest_dataset,
...
-    for row in validation_query_job.result():
-        assert row["count"] == sum(1 for _ in open(test_data_file))
+    expected_num_rows = sum(1 for _ in open(test_data_file))
+    try:
+        bq_wait_for_rows(bq, dest_table, expected_num_rows)
+    except TimeoutError as err:
+        raise AssertionError from err
+
+
+def bq_wait_for_rows(bq_client: bigquery.Client, table: bigquery.Table,
```
jaketf, line 369: Pivoting to this approach has two benefits:
1. table.get requests are free (v.s SELECT COUNT(*) queries which cost / consume slots)
1. speeds up tests ~50%  on average (depending on bq load slot pool latency).

cloud_functions/gcs_event_based_ingest/tests/resources/.gitignore
```diff
@@ -0,0 +1 @@
+hit_data_2020-06-25_09.tsv
```
jaketf, line 1: this doesn't match any files in source tree remove.

ktopcuoglu:
@jaketf Did you ever seen [bqtail project](https://github.com/viant/bqtail).
That is under apache2 license, and has similarities with this solution.

jaketf:
@ktopcuoglu I had not seen bqtail thanks for the link! That tool seems to have some pretty cool features especially with post actions (something I'm very keen to add as a separate cloud function that subscribes to audit logs PS sink to respond to successful BQ job ids (matching some naming convention) and submit the "next" bq job / gcs rm/mv)! 

In general bqtail (as the name suggests) supports incremental ingest of logs to BQ rather well. However, many businesses care about the completeness of a given "partition" of data (e.g. this hour of data in event time should be committed to BQ atomically). The completeness of this partition can only be signaled by the publisher of the data (e.g. upstream hadoop job dropping a _SUCCESS file) which may be delayed. bqtail instead says ingest these files based on some batch window (notably in processing time not event time) as dictated by cloud scheduler. That being said if your publishers don't publish partitions of data in this way bq tail solves a problem of doing that batching for you! It really depends what kind of publishers you are dealing with (hadoop jobs make lots of temp files that could potentially really confuse bqtail). I'd use bqtail if I had many distributed publishers uploading logs to GCS.

the lack of heirarchical configuration might make bqtail tedious to configure for 1000s of tables that have the same configuration (which is often the case in EDW migrations).

Many customers have security requirements around VPC-SC which unfortunately does not support Cloud Scheduler (as it depends on the App Engine Stack). This is how bqtail achieves batching and post actions.

I'm really glad you brought this up because I can add an alternatives section that explains these things and how a user might choose between the two tools based on their use case.

jaketf:
@danieldeleo @ryanmcdowell what's the next step for this PR? I have some minor features I'd like to add in subsequent PRs but would like to get the main tool merged.
- cli tool to help w/ writing configs to appropriate locations in GCS
- cli tool for "backfilling" notifications (e.g. if you have a bucket that already follows this structure w/ existing success files find them and send pubsub notifications for them to support "we started uploading data to GCS before deploying the cloud function") 
- another cloud function / TF module for running "post actions" based on job id completion of a certain prefix (e.g. run pso data validator, run a bq query, move / delete / archive gcs files)

danieldeleo (approved):
lgtm

ryanmcdowell (approved):
LGTM

ryanmcdowell:
Actually just realized this created a new TLD for cloud_functions. Instead I would move this to the tools/ directory under a separate PR given it's a utility.

cloud_functions/gcs_event_based_ingest/Dockerfile.ci
danieldeleo: I was wondering about this! Thanks for the fyi

PR #189, opened 2020-11-11
move cloud functions to tools

jaketf:
per last comment in #182 lets move this to tools


ryanmcdowell:
Since there is only one tool of this type, I would just get rid of the `cloud_functions` directory and move it to the root of tools/. I think that will make the repo easier to navigate. 

If we find there are more tools of this type then we can create a subdirectory at that point.

jaketf:
I plan to add a "bq job post action" cloud function soon that subscribes to audit logs and can run a downstream query, delete / move / archive GCS files or maybe run an "arbitrary" python function (ideally we could use this to run PSO data validator tool for validations that are known to be quick).

In general directory moves break links (which I'd like to add to customer docs). If we foresee more cloud functions being added then IMHO it's best to create that directory now.

ryanmcdowell (approved):

PR #191, opened 2020-11-18
[wip] Pre commit

jaketf:
@danieldeleo this PR introduces the [pre-commit](https://pre-commit.com/) I'd suggest for this repo as we discussed.
It is work in progress as there is a lot of work to be done to get the existing code to comply to the standards imposed by the pre-commit.

As the repo maintiner you'll have to make decisions about whether to ignore existing directories temporarily with exclude or fix all the non-compliant code in this PR. You may also have more opinions about the various flake8 / hadolint / yamllint / pytest configurations / ignore codes. 

R: @ryanmcdowell 
CC: @potiuk @mik-laj who inspired most of this on an earlier pso project

PR #192, opened 2020-11-19
Pre commit only

jaketf:
fixes #191 based on offline discussion will not run on all files will ignore tools dir.


danieldeleo:
Add yamllint-config.yml file to PR

PR #201, opened 2020-11-25
fix: external configs not found in parent dirs

jaketf:
fixes a bug where external table configs were not properly discovered in parent directories and add an integration test for this.


tools/cloud_functions/gcs_event_based_ingest/tests/conftest.py
```diff
@@ -134,6 +134,30 @@ def teardown():
     return data_objs[-1]
 
 
+@pytest.fixture(scope="function")
+@pytest.mark.usefixtures("gcs_bucket", "dest_dataset", "dest_table")
+def gcs_data_under_sub_dirs(request, gcs_bucket, dest_dataset,
```
danieldeleo, line 139: Can you add docstring with brief explanation of test. The test name isn't clear as to what it's testing.

tools/cloud_functions/gcs_event_based_ingest/tests/conftest.py
```diff
@@ -134,6 +134,30 @@ def teardown():
...
+                            dest_table) -> storage.blob.Blob:
+    data_objs = []
```
jaketf, line 141:
```suggestion
    """test that config files in parents are discovered when a success file
    is uploaded in a child directory."""
    data_objs = []
```

jaketf: @danieldeleo this look good? ^

danieldeleo (approved):

PR #224, opened 2021-04-26
Add function degrees(), reverse of existing function radians()

mentin:
degrees is defined as
```
CREATE OR REPLACE FUNCTION fn.degrees(x ANY TYPE) AS (
  180 * x / ACOS(-1)
);

```
this matches existing radians() defined as
```
CREATE OR REPLACE FUNCTION fn.radians(x ANY TYPE) AS (
  ACOS(-1) * x / 180
);
```

### DIFF
--- a/udfs/community/README.md
+++ b/udfs/community/README.md
@@ -14,6 +14,7 @@ SELECT bqutil.fn.int(1.684)
 ## UDFs
 
 * [csv_to_struct](#csv_to_structstrlist-string)
+* [degrees](#degreesx-any-type)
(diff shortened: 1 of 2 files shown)


danieldeleo:
Thanks for the PR! Could you please add some test cases to the [test_cases.yaml](https://github.com/GoogleCloudPlatform/bigquery-utils/blob/master/udfs/community/test_cases.yaml) which verify the UDF behaves as expected.

danieldeleo (approved):

PR #228, opened 2021-05-07
Add default dataset id command line parameter to JMeter templates

danieldeleo:
Add default_dataset_id as a command line parameter which specifies the default datasetId to assume for any unqualified table names in the queries. If not set, all table names in the query string must be qualified in the format 'datasetId.tableId'.


danieldeleo:
@vicenteg fyi

vicenteg (approved):
lgtm

PR #262, opened 2021-08-31
Update bqutil's builder image to include all necessary build components

danieldeleo:
This will accelerate the build process since once build image will be used for all steps. This will also enable developers to test their UDFs using this bqutil public image.


ryanmcdowell (approved):
LGTM

One consideration:
What are the permissions on the image? Can others run this without access to the bqutil project?

danieldeleo:
> LGTM
> 
> One consideration:
> What are the permissions on the image? Can others run this without access to the bqutil project?

The image is public which does allow others to run without access to the project

PR #265, opened 2021-09-02
Updates to bq_ingest

danieldeleo:


tools/cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/common/ordering.py
```diff
@@ -221,7 +319,21 @@ def handle_backlog(
...
+    lock_contents_str = utils.read_gcs_file_if_exists(
+        gcs_client, f"gs://{bkt.name}/{lock_blob.name}")
+    if lock_contents_str:
+        lock_contents: Dict = json.loads(lock_contents_str)
+        if lock_contents:
+            print(
+                json.dumps(
+                    dict(message=f"View lock contents in jsonPayload for"
+                         f" gs://{bkt.name}/{lock_blob.name}",
+                         lock_contents=lock_contents)))
```
franklinWhaite, line 333: Syntax here is similar to that of the functions in the logging.py module. Would it not be slightly more consistent if we had something in the logging.py to manage this? In the same manner as in [line 82](- https://github.com/danieldeleo/bigquery-utils/blob/bq_ingest_updates/tools/cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/common/ordering.py#L82-L83) and [232](- https://github.com/danieldeleo/bigquery-utils/blob/bq_ingest_updates/tools/cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/common/ordering.py#L232-L234)

tools/cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/common/utils.py
```diff
@@ -40,96 +42,157 @@
...
         gcs_client: storage.Client, bq_client: bigquery.Client, gsurl: str,
-        query: str, dest_table_ref: bigquery.TableReference, job_id: str):
+        query: str, job_id: str, table: bigquery.TableReference):
     """Load from query over external table from GCS.
 
     This hinges on a SQL query defined in GCS at _config/*.sql and
-    an external table definition _config/external.json (otherwise will assume
+    an external table definition
+    _config/{constants.BQ_EXTERNAL_TABLE_CONFIG_FILENAME} (otherwise will assume
     CSV external table)
```
franklinWhaite, line 56: Shouldn't this be PARQUET instead of CSV?

danieldeleo: Good catch!

tools/cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/common/ordering.py
danieldeleo: The logging.py holds helper functions for standardized logging that can be used for log filters/routers. This particular instance of a print/log statement is a one-off which is more for debugging than filtering/routing. I do agree we should manage as much in logging.py as possible but I'd prefer to not created a helper function for just this one print statement.

franklinWhaite: Understood, thanks for the explanation. 

I believe the same print statement is also used [here](https://github.com/danieldeleo/bigquery-utils/blob/bq_ingest_updates/tools/cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/common/ordering.py#L122)

Anyhow if think we are ok with leaving these print statements as they are. The filters/routers vs debugging explanations makes enough sense.

tools/cloud_functions/gcs_event_based_ingest/gcs_ocn_bq_ingest/common/utils.py
```diff
@@ -40,96 +42,157 @@
...
+                get_hive_partitioning_source_uri_prefix(
+                    external_table_def["sourceUris"][0])
+        }
     external_config = bigquery.ExternalConfig.from_api_repr(external_table_def)
     job_config = bigquery.QueryJobConfig(
         table_definitions={"temp_ext": external_config}, use_legacy_sql=False)
 
-    # drop partition decorator if present.
-    table_id = dest_table_ref.table_id.split("$")[0]
+    if table:
```
franklinWhaite, line 100: This validation confuses me a bit. (I might be missing something).

I guess we are trying to check if `table is None`.

1) If `table is None` I believe the CF might exit without writing any data to BQ, isn't this worthy of being logged or raising a warning? Is there a case where this is the desired behavior?

2) In this same function, a method is called on the `table` variable previous to this validation. Would that not trigger a an error if `table is None`?

danieldeleo: This is a good catch, I think we can just eliminate the 'If' check.

franklinWhaite: Agree

PR #291, opened 2022-06-09
Fix billing analytics SQL script

Amskee:
**Issue**

BigQuery billing analytics [monthly_invoice_costs_credits.sql](https://github.com/GoogleCloudPlatform/bigquery-utils/blob/master/scripts/billing/sql/monthly_invoice_costs_credits.sql) script uses join to unnest credits and adds them to total cost. This produces duplicate cost rows and incorrect total cost when there is more than one credit applied to the same line item.

image

**Fix**

Join is replaced with subquery to prevent duplication of rows which fixes the total cost.


danieldeleo (approved):
Good catch and thanks for the fix!

PR #293, opened 2022-07-02
Extract Directory Level FROM URL by url_pasrse function

1987yama3:
I add part to extract component from url by url_parse function.
It can extract directory level(DIR1, DIR2, DIR3, DIR4, DIR5).


1987yama3:
@danieldeleo Can you review this issue?

danieldeleo:
This doesn't seem appropriate for the current function. What happens when a user needs the 6th directory?

PR #301, opened 2022-09-11
Patch: enable UDF "url_param" to handle tildes and underscores

sakamossan:
We can use underscore and tilde in URL. I have made them capturable.

> Characters that are allowed in a URI but do not have a reserved
> purpose are called unreserved. These include uppercase and lowercase
> letters, decimal digits, hyphen, period, underscore, and tilde.

- [RFC 3986: Uniform Resource Identifier (URI): Generic Syntax](https://www.rfc-editor.org/rfc/rfc3986#section-2.3)

### DIFF
--- a/udfs/community/README.md
+++ b/udfs/community/README.md
@@ -815,7 +815,7 @@ SELECT bqutil.fn.url_param(
 ```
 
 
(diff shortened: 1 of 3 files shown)


danieldeleo (approved):
Thanks for adding this fix!

PR #309, opened 2022-10-13
typeof() UDF improvement: correctly identify single quoted string

thijs-obj:
Improvement to the `typeof()` UDF: correctly recognize more (all?) strings.

`FORMAT('%T', <string value>)` will return a string with single quotes if the passed string value contains a double quote (`"`) and no single quotes (`'`). `typeof()` would no recognize that as a string, and return `"UNKNOWN"`.

### DIFF
--- a/udfs/community/typeof.sqlx
+++ b/udfs/community/typeof.sqlx
@@ -27,7 +27,7 @@ AS ( (
         WHEN REGEXP_CONTAINS(literal, r'^-?[0-9]*$') THEN 'INT64'
         WHEN REGEXP_CONTAINS(literal, r'^(-?[0-9]+[.e].*|CAST\("([^"]*)" AS FLOAT64\))$') THEN 'FLOAT64'
         WHEN literal IN ('true', 'false') THEN 'BOOL'
-        WHEN literal LIKE '"%' or literal LIKE "'%" THEN 'STRING'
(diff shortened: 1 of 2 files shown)


thijs-obj:
The CLA will take some days to get sorted here.

thijs-obj:
I'd like to request a review of this small PR. Not really clear what the process for that is, so I'll just mention the last few reviewers I saw on other PRs: @danieldeleo @shen-chao @yzhvictor 

Thanks in advance!

udfs/community/typeof.sqlx
```diff
@@ -27,7 +27,7 @@ AS ( (
         WHEN REGEXP_CONTAINS(literal, r'^-?[0-9]*$') THEN 'INT64'
         WHEN REGEXP_CONTAINS(literal, r'^(-?[0-9]+[.e].*|CAST\("([^"]*)" AS FLOAT64\))$') THEN 'FLOAT64'
         WHEN literal IN ('true', 'false') THEN 'BOOL'
-        WHEN literal LIKE '"%' THEN 'STRING'
+        WHEN literal LIKE '"%' or literal LIKE "'%" THEN 'STRING'
```
danieldeleo, line 30:
```suggestion
        WHEN literal LIKE '"%' OR literal LIKE "'%" THEN 'STRING'
```

danieldeleo (approved):
Thanks for the fix!

PR #310, opened 2022-10-13
typeof() UDF improvement: correctly identify single quoted string

thijs-obj:
Improvement to the `typeof()` UDF: correctly recognize more (all?) strings.

`FORMAT('%T', <string value>)` will return a string with single quotes if the passed string value contains a double quote (`"`) and no single quotes (`'`). `typeof()` would no recognize that as a string, and return `"UNKNOWN"`.

This is attempt two. CLA should be OK now


thijs-obj:
Closing this as this didn't fix the CLA issue

PR #334, opened 2022-10-29
Handling Invalid input for  `fn.json_extract_keys`,  `fn.json_extract_values`

takegue:
In BigQuery SQL function behavior,
function typically would return null for invalid arguments, not causing runtime error. JavaScript UDF currently does.

The difference is not useful for users because it gives users extra attention to use JavaScript UDFs

This PR fixes behavior of json_extract_keys, json_extract_values

### DIFF
--- a/udfs/community/json_extract_keys.sqlx
+++ b/udfs/community/json_extract_keys.sqlx
@@ -21,5 +21,18 @@ config { hasOutput: true }
 CREATE OR REPLACE FUNCTION ${self()}(input STRING)
 RETURNS Array<String>
 LANGUAGE js AS """
-  return Object.keys(JSON.parse(input));
+  if(input === null) {
+    return null;
+  }
+
+  let obj;
+  try {
+    obj = JSON.parse(input);
+  } catch {
+    return null;
+  }
+  if(!obj) {
+    return null;
+  }
+  return Object.keys(obj);
 """;
--- a/udfs/community/json_extract_keys.sqlx
+++ b/udfs/community/json_extract_keys.sqlx
@@ -21,18 +21,9 @@ config { hasOutput: true }
 CREATE OR REPLACE FUNCTION ${self()}(input STRING)
 RETURNS Array<String>
 LANGUAGE js AS """
-  if(input === null) {
-    return null;
-  }
-
-  let obj;
   try {
-    obj = JSON.parse(input);
+    return Object.keys(JSON.parse(input));
   } catch {
     return null;
   }
-  if(!obj) {
-    return null;
-  }
-  return Object.keys(obj);
 """;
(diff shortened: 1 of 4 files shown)


udfs/community/json_extract_values.sqlx
```diff
@@ -21,5 +21,17 @@ config { hasOutput: true }
...
+  let obj;
+  try {
+    obj = JSON.parse(input);
+  } catch {
+    return null;
+  }
+  if(!obj) {
+    return null;
+  }
+  return Object.values(obj);
```
danieldeleo, line 36:
```suggestion
  try {
    return Object.values(JSON.parse(input));
  } catch {
    return null;
  }
```

danieldeleo:
Please update README description for UDFs to explain new behavior. 
Also add description to UDF DDL using options parameter

udfs/community/json_extract_keys.sqlx
```diff
@@ -21,5 +21,9 @@ config { hasOutput: true }
 CREATE OR REPLACE FUNCTION ${self()}(input STRING)
 RETURNS Array<String>
 LANGUAGE js AS """
```
danieldeleo, line 23: Add description to DDL: `OPTIONS(description="...UDF description...")`

danieldeleo, line 23: Add description to DDL: `OPTIONS(description="...UDF description...")`

udfs/community/json_extract_values.sqlx
```diff
@@ -21,5 +21,9 @@ config { hasOutput: true }
 CREATE OR REPLACE FUNCTION ${self()}(input STRING)
 RETURNS Array<String>
 LANGUAGE js AS """
```
danieldeleo, line 23: Add description to DDL: `OPTIONS(description="...UDF description...")`

takegue:
@danieldeleo Thanks for your reviews.
I have updated README.md and description. plz check them.

SPTKL:
It seems like `json_extract_values` cannot extract values when the values are objects e.g. 
```sql
SELECT bqutil.fn.json_extract_values(
  '{"foo" : {"cat":1}, "bar": {"dog":2}, "hat": "rat"}'
) AS keys_array
```
would return 
image
Any ideas on how to tackle this?

danieldeleo:
> It seems like `json_extract_values` cannot extract values when the values are objects e.g.
> 
> ```sql
> SELECT bqutil.fn.json_extract_values(
>   '{"foo" : {"cat":1}, "bar": {"dog":2}, "hat": "rat"}'
> ) AS keys_array
> ```
> 
> would return image Any ideas on how to tackle this?

That's correct. This function explicitly declares an output of Array\<String\>. 
Elements in a BigQuery array must all have the same type ([source](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#array_type:~:text=Elements%20in%20an%20array%20must%20share%20the%20same%20type)). 
Therefore you'd have to create a new function which would return an array of STRUCT type and you'd have to handle the logic of converting simple strings to STRUCTs assuming your input JSON has varying types.

PR #339, opened 2022-12-14
Add json_merge UDF function

ismailsimsek:
Combines given two json values and returns combined json.  Attributes from second json are added to first json if missing and replaced if they are exists.

This enables incremental proces key value data(aggregated to json field), and enables appending,updating json field with new data.


ismailsimsek:
> Please could you move your test cases in udf/community/test_cases.js file. Please could you also provide description at udf/community/README.md file.

updated

rajc242 (changes requested):
Please could you add your test cases in udfs/community/test_cases.js file.

PR #345, opened 2023-02-01
Updated daily and hourly utilization to use JOBS_TIMELINE

geoffmc:
Current implementation uses the JOBS table which doesn't account for long running jobs. In the hourly_utilization query, if a job runs for multiple hours, then all of the slot usage gets counted in the first hour. Using JOBS_TIMELINE, this usage gets broken across each hour that it runs.


dashboards/system_tables/sql/daily_utilization.sql
```diff
@@ -28,17 +28,18 @@
   */
 
 SELECT
-  TIMESTAMP_TRUNC(jbo.creation_time, DAY) AS usage_date,
+  TIMESTAMP_TRUNC(jbo.period_start, DAY) AS usage_date,
```
danieldeleo, line 31:
```suggestion
  TIMESTAMP_TRUNC(jbo. job_start_time, DAY) AS usage_date,
```

danieldeleo (approved):

PR #356, opened 2023-05-16
Translation Engine UDF for signed 128bit bitshift.

damian-compilerworks:
UDFs for right/left bit shift of signed 128-bit integer (which is modeled as BIGNUMERIC with 0 scale).


danieldeleo (approved):
Thanks!

PR #364, opened 2023-06-16
[UDF] Add UDF for emulating snowflake split_part

bdw-g:
This adds a function for emulating snowflake split_part. This function has the following semantics:

- The part index is 1-based
- However, index 0 is treated as 1
- Negative numbers are interpreted as indexes from the end of the string

It will help customers migrate from Snowflake to BigQuery.

### DIFF
--- a/udfs/community/test_cases.js
+++ b/udfs/community/test_cases.js
@@ -3009,7 +3009,96 @@ generate_udf_test("cw_period_rdiff", [
         expected_output: `NULL`
     },
 ]);
-
+generate_udf_test("cw_split_part_delimstr_idx", [
+  {
+    inputs: [
+      `"foo bar baz"`,
+      `" "`,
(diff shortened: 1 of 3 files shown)


udfs/community/test_cases.js
```diff
@@ -3009,7 +3009,72 @@ generate_udf_test("cw_period_rdiff", [
         expected_output: `NULL`
     },
 ]);
-
+generate_udf_test("cw_split_part_delimstr_idx", [
```
danieldeleo, line 3012: Please add test case for NULL inputs

danieldeleo (approved):
Will merge after null test cases are added, thanks!

udfs/community/test_cases.js
bdw-g: Done.

PR #369, opened 2023-07-06
Fix edge cases in JavaScript regexp_extract and regexp_extract_all UDFs

mswapnilG:
Fixed following issue with `cw_regexp_extract` and `cw_regexp_extract_all` UDFS:
1. When no match occurs then UDF returns null instead of throwing error
2. When any of the input to UDFs is null then it returns null instead of throwing error.


danieldeleo:
Thanks for the fix! Please also update the comments for the UDFs in the README.md file.

mswapnilG:
@danieldeleo Done

danieldeleo (approved):

PR #373, opened 2023-07-19
Support arbitrary precision input in `cw_convert_base` for supported radix.

paolomorandini:
This change introduces the use of `BigInt` (arbitrary precision integer numbers) in `cw_convert_base`.
`BigInt` will be used for supported radix (2,8,10,16) else the function will default to `parseInt` which can safely represent integer numbers within approximately `+/- 2^53`.


arotenberg-google (changes requested):
Can we just do a [manual conversion](https://stackoverflow.com/questions/55646698/base-36-to-bigint)? At least for the case where we currently fall back to `parseInt`, so that we have identical behavior across all bases and not behavior that works for large magnitude integers for some bases and not others.

danieldeleo (approved):

PR #374, opened 2023-07-26
add alternate largest tables scripts

CYarros10:
alternate scripts that fix the storage size for tables in results and split full_table_id into 3 columns


danieldeleo:
Please make your changes to the existing files instead of creating new ones so that the diff of your changes is clearly shown. Thanks!

PR #375, opened 2023-07-27
update largest tables scripts

CYarros10:
+ expand full table id for greate flexibility when querying table
+ group by fixed for accurate table sizes (previously agg sum for entire project instead of per table)
+ improved readability for group by
+ formatting


danieldeleo (approved):

danieldeleo:
Thanks @CYarros10 !

PR #385, opened 2023-10-03
[UDF] Add strtok function for teradata emulation

bdw-g:
STRTOK in Teradata is a bit special (or rather, it's similar to libc's strtok, and that's a bit special):

- ignores leading delimiters
- ignores duplicate delimiters
- returns NULL for out-of-band fields
- except if the whole string consists of delimiters, and you want the first token, in which case it returns an empty string

### DIFF
--- a/udfs/community/test_cases.js
+++ b/udfs/community/test_cases.js
@@ -2458,6 +2458,91 @@ generate_udf_test("cw_strtok", [
                            STRUCT(CAST(2 AS INT64) AS tokennumber, "1" AS token)])`
     },
 ]);
+generate_udf_test("cw_td_strtok", [
+    {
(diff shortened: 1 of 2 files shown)


udfs/community/test_cases.js
```diff
@@ -2458,6 +2458,91 @@ generate_udf_test("cw_strtok", [
...
+          `";-"`,
+          `3`
+        ],
+        expected_output: `"c"`
+    },
+    {
+        inputs: [
+          `"aa"`,
+          `";"`,
+          `1'`
```
danieldeleo, line 2531:
```suggestion
          `1`
```

danieldeleo (approved):
Lgtm

PR #386, opened 2023-10-04
[UDF] strtok null safety

bdw-g:
Always handle nulls in SQL.


danieldeleo (approved):
Love it! Test all the things! :D

PR #390, opened 2023-12-18
Update queries_grouped_by_hash.sql

afleisc:


scripts/optimization/queries_grouped_by_hash.sql
```diff
@@ -53,10 +53,11 @@ SELECT
   ARRAY_AGG(
     STRUCT(
       bqutil.fn.job_url(project_id || ':us.' || parent_job_id) AS parent_job_url,
-      bqutil.fn.job_url(project_id || ':us.' || job_id) AS job_url
+      bqutil.fn.job_url(project_id || ':us.' || job_id) AS job_url,
+      query as query_text
```
danieldeleo, line 57: I purposely left out the query text here because it doesn't display well when the query is very large and since we're grouping by query hash they won't differ much if at all in this array. I expected users to click the job url to see the query text in the BQ console where it's more nicely formatted and has all the execution details/graph as well. 

Did you have some use-case in mind for adding the query text here?

danieldeleo (changes requested):

scripts/optimization/queries_grouped_by_hash.sql
afleisc: My idea here was to run the SQL anti pattern tool on the query text. I ran a subsequent query to pull the hash ID + query text of one of the jobs in the top 10 into a separate table. I modified the anti pattern tool to take an input BQ table as a parameter.

This doesn't have to be part of this query, but we could add another query just to get the SQL text

danieldeleo (approved):
great use-case, alright let's add it in and let users decide if they want to comment it out or not

PR #410, opened 2024-05-14
Add `MATCH_RECOGNIZE` related UDFs to the community UDFs.

mswapnilG:
Following UDFs are added:
1. `cw_disjoint_partition_by_regexp` - Partitions rows into disjoint segments by matching their sequence with the provided regex pattern
2. `cw_overlapping_partition_by_regexp` - Partitions rows into overlapping segments by matching their sequence with the provided regex pattern

### DIFF
--- a/udfs/community/cw_disjoint_partition_by_regexp.sqlx
+++ b/udfs/community/cw_disjoint_partition_by_regexp.sqlx
@@ -15,8 +15,9 @@ config { hasOutput: true }
  * limitations under the License.
  */
 
-/*
- Returns disjoint matches for the regex with the sequence of rows encoded in
+CREATE OR REPLACE FUNCTION ${self()}(firstRn INT64, haystack STRING, regex STRING)
+  RETURNS ARRAY<INT64> 
+  OPTIONS(description="""Returns disjoint matches for the regex with the sequence of rows encoded in
  custom format. The expected input haystack format is
  "row-1@row-number-1#row-2@row-number-2# ... #row-n@row-number-n#".
  If the regex matches then, it returns the rows matched by the given pattern
@@ -39,9 +40,8 @@ config { hasOutput: true }
  3             "A@1#A@2#B@3#A@4#B@5#"  []
  4             "A@1#A@2#B@3#A@4#B@5#"  [4, 5]
  5             "A@1#A@2#B@3#A@4#B@5#"  []
- */
-CREATE OR REPLACE FUNCTION ${self()}(firstRn INT64, haystack STRING, regex STRING)
-  RETURNS ARRAY<INT64> AS (
+""")
+  AS (
     (WITH t AS (
         SELECT
           MIN(ARRAY_LENGTH(REGEXP_EXTRACT_ALL(m, '@\\d+#'))) AS n_rows
--- a/udfs/community/cw_disjoint_partition_by_regexp.sqlx
+++ b/udfs/community/cw_disjoint_partition_by_regexp.sqlx
@@ -1,6 +1,6 @@
 config { hasOutput: true }
 /*
- * Copyright 2023 Google LLC
+ * Copyright 2024 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/udfs/community/cw_disjoint_partition_by_regexp.sqlx
+++ b/udfs/community/cw_disjoint_partition_by_regexp.sqlx
@@ -1,0 +1,52 @@
+config { hasOutput: true }
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


udfs/community/cw_disjoint_partition_by_regexp.sqlx
```diff
@@ -0,0 +1,52 @@
...
+ increasing values, then we get corresponding outputs as follows:
+ firstRn       haystack                output
+ 1             "A@1#A@2#B@3#A@4#B@5#"  [1, 2, 3]
+ 2             "A@1#A@2#B@3#A@4#B@5#"  []
+ 3             "A@1#A@2#B@3#A@4#B@5#"  []
+ 4             "A@1#A@2#B@3#A@4#B@5#"  [4, 5]
+ 5             "A@1#A@2#B@3#A@4#B@5#"  []
+ */
+CREATE OR REPLACE FUNCTION ${self()}(firstRn INT64, haystack STRING, regex STRING)
+  RETURNS ARRAY<INT64> AS (
```
danieldeleo, line 44:
```suggestion
CREATE OR REPLACE FUNCTION ${self()}(firstRn INT64, haystack STRING, regex STRING)
  RETURNS ARRAY<INT64> 
  OPTIONS(description="""Returns disjoint matches for the regex with the sequence of rows encoded in
 custom format. The expected input haystack format is
 "row-1@row-number-1#row-2@row-number-2# ... #row-n@row-number-n#".
 If the regex matches then, it returns the rows matched by the given pattern
 if the firstRn is the first row in the given matched subsequence. e.g. if
 firstRn is 4,  haystack is "A@1#A@2#B@3#A@4#B@5#" and regex is
 "(?:A@\\d+#)+(?:B@\\d+#)" then it will return [4, 5] since matched subsequence
 is "A@4#B@5#" starts at firstRn 4, however if you have input firstRn is 5 with
 the same input then it will return an empty array since row number 5 is part
 of "A@4#B@5#" but it is not the first row in the subsequence.

 By repeatedly calling cw_disjoint_partition_by_regexp with increasing values of
 firstRn and keeping haystack constant, this UDF effectively returns disjoint
 sets of row numbers that match the given regex.

 Continuing the above example, if we call this UDF with firstRn having
 increasing values, then we get corresponding outputs as follows:
 firstRn       haystack                output
 1             "A@1#A@2#B@3#A@4#B@5#"  [1, 2, 3]
 2             "A@1#A@2#B@3#A@4#B@5#"  []
 3             "A@1#A@2#B@3#A@4#B@5#"  []
 4             "A@1#A@2#B@3#A@4#B@5#"  [4, 5]
 5             "A@1#A@2#B@3#A@4#B@5#"  []
"""
  AS (
```

udfs/community/cw_overlapping_partition_by_regexp.sqlx
```diff
@@ -0,0 +1,49 @@
...
+ get corresponding outputs as follows:
+ firstRn    haystack                output
+ 1          "A@1#A@2#B@3#A@4#B@5#"  [1, 2, 3]
+ 2          "A@2#B@3#A@4#B@5#"      [2, 3]
+ 3          "B@3#A@4#B@5#"          []
+ 4          "A@4#B@5#"              [4, 5]
+ 5          "B@5#"                  []
+ */
+CREATE OR REPLACE FUNCTION ${self()}(firstRn INT64, haystack STRING, regex STRING)
+  RETURNS ARRAY<INT64> AS (
```
danieldeleo, line 44:
```suggestion
CREATE OR REPLACE FUNCTION ${self()}(firstRn INT64, haystack STRING, regex STRING)
  RETURNS ARRAY<INT64> 
  OPTIONS(description="""Returns overlapping matches for the regex with the sequence of rows encoded in
 custom format. The expected input haystack format is
 "row-1@row-number-1#row-2@row-number-2# ... #row-n@row-number-n#".
 If the regex matches, then it returns all rows matched at the prefix of the
 haystack,  e.g. if firstRn is 1, firstRn is "A@1#A@2#B@3#A@4#B@5#" and regex
 is "(?:A@\\d+#)+(?:B@\\d+#)" then it will return [1, 2, 3] as output since
 matched subsequence is "A@1#A@2#B@3#". If haystack is "B@3#A@4#B@5#" for the
 same regex it will return an empty array since regex does not match at the
 prefix of the haystack.

 By repeatedly calling cw_overlapping_partition_by_regexp with increasing values
 of firstRn and a substring of sequence starting at firstRn, this UDF
 effectively returns overlapping sets of row numbers that match the given regex.

 Continuing the above example, if we call this UDF with firstRn having
 increasing values, and haystack having substring starting with firstRn, then we
 get corresponding outputs as follows:
 firstRn    haystack                output
 1          "A@1#A@2#B@3#A@4#B@5#"  [1, 2, 3]
 2          "A@2#B@3#A@4#B@5#"      [2, 3]
 3          "B@3#A@4#B@5#"          []
 4          "A@4#B@5#"              [4, 5]
 5          "B@5#"                  []
"""
)
  AS (
```

udfs/community/cw_disjoint_partition_by_regexp.sqlx
```diff
@@ -0,0 +1,52 @@
+config { hasOutput: true }
+/*
+ * Copyright 2023 Google LLC
```
danieldeleo, line 3:
```suggestion
 * Copyright 2024 Google LLC
```

udfs/community/cw_overlapping_partition_by_regexp.sqlx
```diff
@@ -0,0 +1,49 @@
+config { hasOutput: true }
+/*
+ * Copyright 2023 Google LLC
```
danieldeleo, line 3:
```suggestion
 * Copyright 2024 Google LLC
```

danieldeleo (approved):

PR #417, opened 2024-05-29
fixes jaccard bug

### LINKED ISSUE #383: Bug: Jaccard UDF giving incorrect results in some cases
I found a bug in the the Jaccard distance function https://github.com/GoogleCloudPlatform/bigquery-utils/blob/master/udfs/community/jaccard.sqlx (the function below is a copy of the Jaccard UDF, cast as a TEMP FUNCTION)

```
CREATE TEMP FUNCTION Jaccard(a STRING, b STRING)
RETURNS FLOAT64
LANGUAGE js AS r"""
  let intersectSize = 0;

  if(a && b) {
    // Split word into characters, and make characters unique
    const sa = [...new Set(String(a).split(''))]
    const sb = [...new Set(String(b).split(''))]

    const la = (sa.length > sb.length) ? sa.length : sb.length
    const lb = (sa.length > sb.length) ? sb.length : sa.length

    for(let i = 0; i < la; i++) {
      for(let j = 0; j < lb; j++) {
        intersectSize += (sa[i] === sb[j]) ? 1 : 0
      }
    }

    // Compute Jaccard distance
    const union = (sa.length + sb.length) - intersectSize
    return parseFloat(((intersectSize / union) * 100 + Number.EPSILON) / 100).toFixed(2)
  } else {
    return 0.0
  }
""";

with datas as (
select 
'12' as a,
'132' as b
)

select
 Jaccard(a, b)
from datas
```
This should give 2/3 = 0.667 (since here there the intersection is size 2, and the union is size 3), but instead gives 0.25.

I believe the bug can be fixed by replacing 
```
    const la = (sa.length > sb.length) ? sa.length : sb.length
    const lb = (sa.length > sb.length) ? sb.length : sa.length
```
with 
```
    const la = sa.length
    const lb = sb.length
```
With the current version of these lengths, the array indices become out of bounds.

afleisc:
Closes #383


afleisc:
```
const la = (sa.length > sb.length) ? sa.length : sb.length
const lb = (sa.length > sb.length) ? sb.length : sa.length
```

was added as an optimization to the double for loop but introduces a bug.

The incorrect loop limits could cause the intersection count (intersectSize) to be inaccurate, as not all possible character pairs were being compared.

danieldeleo:
Sorry just getting to this, can you pull the latest to resolve merge conflicts

danieldeleo (approved):

PR #438, opened 2024-08-05
[Sketch] Make merge more defensive for calling order

bdw-g:
The existing implementations assume that merge() only ever happens after deserialization. This assumption is invalid, and the implementation should be correct for merge() directly after aggregation.

### DIFF
--- a/udfs/community/tuple_sketch_int64.sqlx
+++ b/udfs/community/tuple_sketch_int64.sqlx
@@ -147,12 +147,10 @@ export function merge(state, other_state) {
     state.union = Module._tuple_union_initialize(state.lg_k);
   }
 
-  if (state.sketch || other_state.sketch) {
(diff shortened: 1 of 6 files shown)


udfs/community/tuple_sketch_int64.sqlx
```diff
@@ -147,12 +147,10 @@ export function merge(state, other_state) {
     state.union = Module._tuple_union_initialize(state.lg_k);
   }
 
-  if (state.sketch || other_state.sketch) {
-    throw new Error("update_sketch not expected during merge()");
-  }
-
-  if (other_state.union) {
-    throw new Error("other_state should not have union during merge()");
+  if (state.sketch)
```
danieldeleo, line 150:
```suggestion
  if (state.sketch) {
```

danieldeleo (approved):

PR #450, opened 2024-09-04
fix: honor `offset` when `occurence` is 0

shobsi:
```sql
SELECT bqutil.fn.cw_regexp_replace_4('abaa', 'a', 'A', 1)
```

Current behavior:
```
AbAA
```

Expected behavior:
```
abAA
```

### DIFF
--- a/udfs/community/test_cases.js
+++ b/udfs/community/test_cases.js
@@ -1449,6 +1449,24 @@ generate_udf_test("cw_regexp_replace_4", [
     expected_output: `"TestCad$123456"`,
   },
 ]);
+generate_udf_test("cw_regexp_replace_4", [
+  {
+    inputs: [`"TestStr123456Str"`, `"Str"`, `"Cad$"`, `CAST(1 AS INT64)`],
+    expected_output: `"TestCad$123456Cad$"`,
+  },
+]);
+generate_udf_test("cw_regexp_replace_4", [
+  {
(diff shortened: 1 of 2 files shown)


udfs/community/test_cases.js
```diff
@@ -1449,6 +1449,24 @@ generate_udf_test("cw_regexp_replace_4", [
...
+    inputs: [`"abaa"`, `"a"`, `"A"`, `CAST(1 AS INT64)`],
+    expected_output: `"AbAA"`,
+  },
+]);
+generate_udf_test("cw_regexp_replace_4", [
+  {
+    inputs: [`"abaa"`, `"a"`, `"A"`, `CAST(2 AS INT64)`],
+    expected_output: `"abAA"`,
+  },
+]);
```
danieldeleo, line 1469: This test is failing with the following error:
expected "abAA", but saw "bAA".

danieldeleo (changes requested):

udfs/community/test_cases.js
shobsi: Oh, thanks, sent a patch. Is there a way I can see the test failure logs myself? Is it in one of those 45 checks?

danieldeleo: Yes it's one of those 45 checks (the one named "bigquery-utils-push-to-pr"). Here's the check that was failing: https://github.com/GoogleCloudPlatform/bigquery-utils/pull/450/checks?check_run_id=30891233416

danieldeleo (approved):

Issue #454, opened 2024-09-26
Add optimization script for identifying datasets which would benefit from compressed storage billing and then another script to ALTER them to compressed billing


danieldeleo:
Use [this script](https://cloud.google.com/bigquery/docs/information-schema-table-storage#example_2) as a starting point, but needs to be modified to run over an entire organization

PR #456, opened 2024-09-26
storage billing model ddl optimization script

afleisc:
- The user will be responsible for adjusting the pricing according to their region, and whether or not they want to enforce a smaller time travel period (default is keeping time travel the same), as well as a few other optional parameters found within the script itself


afleisc:
closes #454

danieldeleo (approved):

PR #457, opened 2024-09-27
Fix embedding batch table path to be wrapped in backticks

poffdeluxe:
I was making use of the very awesome `bqml_generate_embeddings` stored procedure and noticed that for the target table, the full path wasn't wrapped in backticks in one spot which causes the call to fail if the project name or table name have odd characters (like a `-` for example)


danieldeleo (approved):

danieldeleo:
Thanks for the fix, @poffdeluxe !

PR #470, opened 2025-01-10
Fix typos in community UDF examples

olivierjeunen:
The current examples do not work. This is due to the project name being "bqutils" instead of "bqutil".
Fixing the typo means that the examples can now be copy-pasted and expected to work.


danieldeleo (approved):
Thanks for catching this!

olivierjeunen:
You're welcome!
Completed the CLA so looks like this should be good to merge. Thanks!

PR #471, opened 2025-01-14
Refactor queries_grouped_by_hash and add 4 additional Optimization Scripts

afleisc:
Some customers have a greater focus on duration of queries rather than total slots. The queries_grouped_by_hash_project/org_duration scripts break each query down by duration percentile rather than focusing solely on slot time which can save time rather than modifying a previous script.

Added two additional scripts to gain insights into customers who use BI Engine. One focuses on the duration of each BI Engine mode so they can monitor whether or not BI Engine is helping. The other focuses on why BI Engine was disabled so they can understand why BI Engine may not have been used. 

Refactored queries_grouped_by_hash to add ordering and limit clauses. The current comments in the scripts don't align with the script logic. 

Also updated the README to reflect some of the latest changes to queries_grouped_by_hash. There were broken links because we broke queries_grouped_by_hash into two separate scripts for project and organization.

### DIFF
--- a/scripts/optimization/bi_engine_mode_duration.sql
+++ b/scripts/optimization/bi_engine_mode_duration.sql
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


scripts/optimization/bi_engine_mode_duration.sql
```diff
@@ -0,0 +1,64 @@
...
+                 AND jbo.end_time > jbo.start_time
+                 AND jbo.error_result IS NULL
+                 AND jbo.statement_type != 'SCRIPT'                
+         )
+ )
+GROUP BY
+ 1,
+ 2
+ ORDER BY day, bi_engine_mode ASC;
+
```
danieldeleo, line 64:
```suggestion

```

scripts/optimization/queries_grouped_by_hash_org_duration.sql
danieldeleo, line 1: remove this file and simply add duration columns to the existing queries_grouped_by_hash_org.sql script

scripts/optimization/queries_grouped_by_hash_project_duration.sql
danieldeleo, line 1: remove this file and simply add duration columns to the existing queries_grouped_by_hash_project.sql script

danieldeleo (changes requested):

PR #492, opened 2025-05-02
use the latest release apache-datasketches-bigquery-1.1.1

AlexanderSaydakov:
Release notes here:
https://github.com/apache/datasketches-bigquery/releases/tag/1.1.0


danieldeleo:
@AlexanderSaydakov https://github.com/apache/datasketches-bigquery/pull/142 introduced breaking changes to the build process for this repo. Can you revert that PR and instead specify that JS_BUCKET values must not end in forward slash? You can have JS_BUCKET specify a path to a folder but you  just need to leave out the last forward slash.

AlexanderSaydakov:
I think I tried that and it did not work, but the modified Makefiles worked for both cases (bucket and folder). I will revisit this.

danieldeleo:
> I think I tried that and it did not work, but the modified Makefiles worked for both cases (bucket and folder). I will revisit this.

The build pipeline for datasketches in bigquery-utils repo passes a folder to the JS_BUCKET variable so I'm certain it works with folders (leaving out the trailing forward slash). Let me know if you can reproduce the issue you saw that led to the PR, I can help debug.

AlexanderSaydakov:
I tested the makefiles when I made the change and re-tested now with JS_BUCKET not having the trailing slash in both cases: gs://bucket and gs://bucket/folder. Works for me.

danieldeleo:
> I tested the makefiles when I made the change and re-tested now with JS_BUCKET not having the trailing slash in both cases: gs://bucket and gs://bucket/folder. Works for me.

You'll need to now revert the changes from https://github.com/apache/datasketches-bigquery/pull/142 in order to publish via this repo

AlexanderSaydakov:
Could you clarify why do you think we need to revert the change? I have just re-tested the release 1.1.0 setting JS_BUCKET=gs;//bucketname and JS_BUCKET=gs://bucketname/foldername (without slash at the end). Both ways work.

danieldeleo:
> Could you clarify why do you think we need to revert the change? I have just re-tested the release 1.1.0 setting JS_BUCKET=gs://bucketname and JS_BUCKET=gs://bucketname/foldername (without slash at the end). Both ways work.

[PR#142](https://github.com/apache/datasketches-bigquery/pull/142) removed the forward slash from the following command:

`gcloud storage cp $$file $(JS_BUCKET)/` 

which results in cp command thinking that the gs:// path is to a file instead of a folder. 
If I pass JS_BUCKET=gs://my_bucket/my_folder, then every file will be written to a file named "my_folder" (overwriting the previous file too), instead of every file being written into a folder named "my_folder".

AlexanderSaydakov:
I guess that in your case the specified folder 2eaa735 does not exist. Is that right? Do you expect the copy command to create the folder?
I tested deploying into existing folder, and it works for me.

danieldeleo:
> I guess that in your case the specified folder "2eaa735" does not exist. Is that right? Do you expect the copy command to create the folder? I tested deploying into existing folder, and it works for me.

Correct, the specified folder doesn't exist nor should it need to since folders aren't actually real objects by default in Cloud Storage (see https://cloud.google.com/storage/docs/objects#flat-namespace for more details).

AlexanderSaydakov:
Tests pass for me and for several others who voted on the release.

danieldeleo:
> Tests pass for me and for several others who voted on the release.

How are you running the tests?

danieldeleo:
> I am running 'make', 'make install' and 'make test'

I ran these exact commands for tag v1.1.1 and got the same failures.

AlexanderSaydakov:
I am not sure how is this possible

danieldeleo (approved):

PR #493, opened 2025-05-14
feat: adds `gunzip` UDF

plaflamme:
Uses the [`pako`](https://www.npmjs.com/package/pako) js library that implements deflate in pure js. Also relies on [`base64-js`](https://www.npmjs.com/package/js-base64) to deal with BQ's `BYTES` contract which is through base64-encoded strings.


udfs/community/gunzip.sqlx
```diff
@@ -45,7 +45,7 @@ if (gzipped == null) {
   return null;
 }
 
-const bytes = base64_js.toUint8Array(gzipped);
+const bytes = base64_js.toByteArray(gzipped);
```
plaflamme, line 48: Looks like I wanted to add [`js-base64`](https://www.npmjs.com/package/js-base64), not [`base64-js`](https://www.npmjs.com/package/base64-js) 🤦 

Both seem heavily used, so I guess either one will do...

udfs/community/gunzip.sqlx
```diff
@@ -0,0 +1,51 @@
...
+SELECT CAST(
+  bqutil.fn.gunzip(
+    FROM_BASE64("H4sIACYJJWgAA8tIzcnJVyjPL8pJ4QIALTsIrwwAAAA=")
+  )
+  AS STRING
+)
+
+/*
+ * The query result will be the following:
+ * "hello world"
```
danieldeleo, line 40:
```suggestion
 * "hello world\n"
```

udfs/community/test_cases.js
```diff
@@ -3752,3 +3752,14 @@ generate_udf_test("cw_xml_extract", [
     expected_output: `"xy"`,
   },
 ]);
+
+generate_udf_test("gunzip", [
+  {
+    inputs: [`FROM_BASE64("H4sIACYJJWgAA8tIzcnJVyjPL8pJ4QIALTsIrwwAAAA=")`],
+    expected_output: `CAST("hello world" AS BYTES)||FROM_BASE64("Cg==")`,
```
plaflamme, line 3759: Thanks, I wasn't sure what I should put here. I left the value as it appears in the BQ console, but it makes sense that we can use SQL here as well, I even used that in for the `NULL` case!

That said, I just realized that I did `echo "hello world" | gzip | base64` which is why there's a newline. We should probably just put the value without a newline: `H4sIABDwJWgAA8tIzcnJVyjPL8pJAQCFEUoNCwAAAA==` instead (`echo -n "hello world"`). So that the expected output is just `CAST("hello world" AS BYTES)`.

udfs/community/gunzip.sqlx
danieldeleo: GitHub repo for base64-js shows >20M  "Used by" while js-base64 shows ~5M "Used by".
I haven't benchmarked them against each other so just going by popularity, we can keep base64-js. 
Feel free to benchmark them and pick the fastest performing one.

udfs/community/test_cases.js
danieldeleo: I agree, please update to remove the newline from the example and unit test. Thanks!

plaflamme: Done in https://github.com/GoogleCloudPlatform/bigquery-utils/pull/493/commits/a8a6b9de3dde81e3b1a9ce0545c0ee7a9e3b4d03

Thanks for all the fixes!

udfs/community/README.md
```diff
@@ -132,6 +132,7 @@ SELECT bqutil.fn.int(1.684)
 * [get_array_value](#get_array_valuek-string-arr-any-type)
 * [getbit](#getbittarget_arg-int64-target_bit_arg-int64)
 * [get_value](#get_valuek-string-arr-any-type)
+* [gunzip](#gunzip)
```
danieldeleo:
```suggestion
* [gunzip](#gunzipgzipped-bytes)
```

udfs/community/README.md
```diff
@@ -1427,6 +1428,19 @@ results:
 | bbb | aaa | ccc |
 
 
+### [gunzip(gzipped BYTES)](gunzip.sqlx)
+Given compressed BYTES using the `DEFLATE` (a.k.a. `gzip`) algorithm, this method will return the decompressed value as BYTES.
+
+```sql
+SELECT CAST(bqutil.fn.gunzip(FROM_BASE64("H4sIAF4OJWgAA8tIzcnJVyjPL8pJ4QIALTsIrwwAAAA=")) AS STRING)
```
danieldeleo:
```suggestion
SELECT CAST(bqutil.fn.gunzip(FROM_BASE64("H4sIAOL4JWgAA8tIzcnJVyjPL8pJAQCFEUoNCwAAAA==")) AS STRING)
```

danieldeleo (approved):
Thanks for contributing this UDF!

PR #498, opened 2025-06-10
add UDF parse_timestamp

mingjialiu:
native parse_timestamp function parses fraction seconds as part of second using formt `E#S`. But other dialects like oracle sql could parse fraction second independently as `F`. So we need this UDF for that need in BQ translation. 

example:
parse_timestamp("21.234", "E3S")  can be parsed by BQ,  but parse_timestamp("21234", "SSE3F") can not be parsed by BQ

### DIFF
--- a/udfs/community/README.md
+++ b/udfs/community/README.md
@@ -155,7 +155,7 @@ SELECT bqutil.fn.int(1.684)
 * [nlp_compromise_number](#nlp_compromise_numberstr-string)
 * [nlp_compromise_people](#nlp_compromise_peoplestr-string)
 * [normal_cdf](#normal_cdfx-float64-mean-float64-stdev-float64)
-* [parse_timestamp_python](#parse_timestamp_python-f-string-s-string)
+* [parse_timestamp](#parse_timestamp-f-string-s-string)
 * [percentage_change](#percentage_changeval1-float64-val2-float64)
 * [percentage_difference](#percentage_differenceval1-float64-val2-float64)
 * [pi](#pi)
@@ -2313,11 +2313,11 @@ results:
 -----
 
 
-### [parse_timestamp_python(f STRING, s STRING)](parse_timestamp.sqlx)
+### [parse_timestamp(f STRING, s STRING)](parse_timestamp.sqlx)
 Parses a timestamp string according to a specified format string. Returns a TIMESTAMP value.
 
 ```sql
-SELECT bqutil.fn.parse_timestamp_python('%Y-%m-%d %H:%M:%S', '2024-03-20 14:30:00');
+SELECT bqutil.fn.parse_timestamp('%Y-%m-%d %H:%M:%S', '2024-03-20 14:30:00');
 
 2024-03-20 14:30:00 UTC
 ```
--- a/udfs/community/parse_timestamp.sqlx
+++ b/udfs/community/parse_timestamp.sqlx
@@ -15,10 +15,11 @@ config { hasOutput: true }
  * limitations under the License.
  */
 
+-- This UDF is added because BigQuery's parse_timestamp does not support the %F format code for fractional seconds, while this UDF does.
 -- Parses a string into a TIMESTAMP using the specified format string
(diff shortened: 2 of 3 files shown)


udfs/community/parse_timestamp.sqlx
```diff
@@ -0,0 +1,31 @@
...
+AS r"""
+from datetime import datetime
+
+
+def do_parse_timestamp(f, s):
+    try:
+        return datetime.strptime(s, f)
+    except ValueError:
+        return None
+""";
```
danieldeleo, line 31:
```suggestion
""";

```

udfs/community/parse_timestamp.sqlx
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- Safely allows greater than comparison on numbers of any magnitude
+CREATE OR REPLACE FUNCTION ${self()}(f STRING, s STRING)
```
danieldeleo, line 19: Thoughts on incorporating the word "python" into the UDF name so that users know it's a python parsing function?

udfs/community/parse_timestamp.sqlx
danieldeleo, line 1: Rename file to match new name

danieldeleo (changes requested):

udfs/community/README.md
```diff
@@ -2316,6 +2312,27 @@ results:
 | -0.4353361094588436 | 0.5300416418798544 | 0.632366563565354 |
 -----
 
+
+### [parse_timestamp_python(f STRING, s STRING)](parse_timestamp.sqlx)
```
mswapnilG, line 2316: Can we keep the neutral name instead of adding the `_python` suffix? UDFs in this folder contain a mix of sql, javascript and python UDFs and they don't append `_sql`, `_js` or `_python`, etc. If you want to differentiate with BQ `parse_timestamp` then maybe we can do something like `parse_timestamp_with_microseconds`?

udfs/community/parse_timestamp_python.sqlx
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- Parses a string into a TIMESTAMP using the specified format string
+CREATE OR REPLACE FUNCTION ${self()}(f STRING, s STRING)
+RETURNS TIMESTAMP LANGUAGE python
+OPTIONS (entry_point='do_parse_timestamp', runtime_version='python-3.11', description='Parses a timestamp string according to a specified format string using Python\'s datetime.strptime(). Returns a TIMESTAMP value or NULL if parsing fails. Supports common format codes such as %Y, %m, %d, %H, %M, %S, and %F for fractional seconds.')
```
mswapnilG, line 21: > using Python\'s datetime.strptime()

This is implementation detail; Instead of this maybe we can explain briefly that we are adding this because BQ `parse_timestamp` doesn't parse `%F` and this UDF does that. 

If we replace this with JS UDF later on (due to performance reason, e.g.), we don't have to change anything except the UDF implementation.

udfs/community/README.md
danieldeleo: I'm good with adding a different suffix, but `with_microseconds` isn't the best one because BQ's parse_timestamp does support parsing microseconds, just in a certain format

udfs/community/parse_timestamp_python.sqlx
mingjialiu: sounds good. Agree that we should add a JS implementation if customers complain about performance and engine team still hasn't support the feature.

PR #499, opened 2025-06-11
feat: add multimodal(mm).exif function

GarrettWu:
The function deployment needs connections ${BQ_LOCATION}.multimodal-udf-connection to be created, which is used for Python udf to access networks. The connections need to be shared publicly for callers to use. No connection SA permissions needed. 

To run the tests, another connection ${BQ_LOCATION}.bigframes-default-connection is needed. Which is a user provided connection to create ObjectRef and access GCS files. storage.ObjectUser role to the GCS bucket is needed for the connection SA. It's user data, nothing related to the function.

I'm reusing the JS_BUCKET for test data. Since there need to be a user bucket to upload and run test against. I feel specifying another bucket is redundant. But now the variable name isn't accurate. Lemme know if we want to rename the variable or need another bucket.

The function "exif" is the one we want users to call. "exif_udf" is a helper, so I didn't include it in the README.

Tests are passing locally with bigframes-dev project. screen/5ZNDCSCfrkURoT8


udfs/multimodal/README.MD
```diff
@@ -0,0 +1,27 @@
+# Multimodal UDFs
+
+This directory contains [user-defined functions](https://cloud.google.com/bigquery/docs/reference/standard-sql/user-defined-functions) of mutimodal data.
+Which deal with [ObjectRef](https://cloud.google.com/bigquery/docs/reference/standard-sql/objectref_functions) of unstructured data in GCS. 
+Each UDF within thisdirectory will be automatically synchronized to the `bqutil` project within the
+`mm` dataset for reference in queries.
+
+## UDFs
+
+* [exif](#exif)
```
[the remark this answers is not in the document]
GarrettWu: I'm afraid that naming doesn't comply with bq-utils' conversions. Functions are named by what they do instead of implementations or dependencies. Also bq-utils only keeps the latest function version (no guarantee on function implementation or dependencies). @danieldeleo @afleisc to comment.

Yes we can ping dependencies versions. Updating.

udfs/community/test_cases.js
```diff
@@ -3763,3 +3763,17 @@ generate_udf_test("gunzip", [
     expected_output: `CAST(NULL AS BYTES)`,
   },
 ]);
+
+generate_udf_test("exif_udf", [
+  {
+    inputs: [`TO_JSON_STRING(OBJ.GET_ACCESS_URL(OBJ.MAKE_REF("${dataform.projectConfig.vars.gcsBucket}/test_data/images_exif/test_image_exif.jpg", "${dataform.projectConfig.defaultLocation}.bigframes-default-connection"), "R"))`],
```
danieldeleo, line 3769: @GarrettWu I understand that users will have to pass their own connection but for unit testing do you have any objections to just re-using multimodal-udf-connection to reduce the amount of connections we need to maintain in our ci/cd infra?

GarrettWu: multimodal-udf-connection is used for UDFs, it has to be shared publicly. And for safety purpose, it doesn't need specific permissions. I was trying to maintain 2 different sets of permissions for those.

1. UDF connection (multimodal-udf-connection): shared publicly, no permissions.
2. User connection (bigframes-default-connection): no sharing, has permissions to project GCS.

If we want to use a single connection, it will work. But we need to be cautious that we don't give users the permissions that we aren't intended to.

danieldeleo: Agreed, the reason I'm ok consolidating is that the GCS bucket is already publicly viewable (required since javascript libraries for JS UDFs are hosted in there). So for unit tests I actually didn't need to grant multimodal-udf-connection any extra permissions since allAuthenticatedUsers can already access the bucket which holds the test_data/ folder

GarrettWu: Is it publicly writable? Later we will have functions to write to bucket as well.

danieldeleo (approved):

PR #539, opened 2026-05-29
Add emulation function for Snowflake's OBJECT_AGG()

Hackerpilot:


danieldeleo (changes requested):
@Hackerpilot I think it'd be good to add unit tests that confirm it behaves exactly the same as snowflake:

Usage notes:

- Input tuples with NULL key and/or value are ignored.
- Duplicate keys within a group result in a Duplicate field key 'key' error.
- The DISTINCT keyword is supported, but it only filters out duplicate rows where both key and value are equal.

Hackerpilot:
* "When a JavaScript UDAF is called with the IGNORE NULLS argument, BigQuery automatically skips over rows for which any aggregate argument evaluates to NULL. "
* This case is handled, but our testing functions currently don't have a way to specify that a function is expected to result in an error.
* This is now handled by a `NOT AGGREGATE` argument (which will be added automatically by the translation engine).

danieldeleo (approved):
Thanks for the changes!

